PR #8, opened 2022-02-12
Add python files to run the  post process data

weizmannk:


weizmannk:
I just add the python's files on the post-processing directory,

mcoughlin:
Did you run pre-commit install and check for linting? @weizmannk

weizmannk:
*Yes now, normally everything will be ok, I check each step  one by one. 
As long as the data (GWdata, EoS, and priors ) are placed in the right place, it will work.

*To upload the EoS data and priors ( without GWdata) in nmma  this weight is approximately 14 MB, and 7MB if I compress eos_sorted file , 
in fact it's the last one which  is weighted ( 13 MB without compress that, and 6.6MB with ) .

example_files/post-processing/R14_trend_generate.py
```diff
@@ -0,0 +1,119 @@
...
+def reweight_to_flat_mass_prior(df):
+    total_mass = bilby.gw.conversion.chirp_mass_and_mass_ratio_to_total_mass(df.chirp_mass, df.mass_ratio)
+    m1 = total_mass / (1. + df.mass_ratio)
+    jacobian = m1 * m1 / df.chirp_mass
+    df_new = df.sample(frac=0.3, weights=jacobian, random_state=42)
+    return df_new
+
+
+# user specify input
+pdet_Mmax_path = '../../example_files/eos/pdet_of_Mmax.dat'
```
mcoughlin, line 20: I think these should be command line arguments.

example_files/post-processing/R14_trend_plot.py
```diff
@@ -0,0 +1,62 @@
...
+          'text.usetex': True,
+          'font.family': 'Times New Roman',
+          'figure.figsize': fig_size}
+matplotlib.rcParams.update(params)
+
+# user specify input
+R14_true = 11.55
+label = 'ZTF'
+
+data_GWEM = pd.read_csv('../../output/Figures/GW_EM_R14trend_{0}.dat'.format(label), header=0, delimiter=' ')
```
mcoughlin, line 29: More command line arguments (and consider specifying all of thse at the top of the script).

example_files/post-processing/R14_trend_plot.py
```diff
@@ -0,0 +1,62 @@
...
+
+data_GW = pd.read_csv('../../example_files/GW_R14trend.dat', header=0, delimiter=' ')
+
+fig = plt.figure()
```
mcoughlin, line 33: Can we add an example plot to the docs?

mcoughlin (changes requested):
It's looking good. A few suggestions.

PR #11, opened 2022-02-14
R14 trend generate

weizmannk:


mcoughlin:
@weizmannk looking better... you should open a PR with just the setup.py and your new files in there right? You don't need these others.

weizmannk:
Yes , i was just to be more precisely.

mcoughlin:
@weizmannk just merged an edited version of the files. lemme know what you think.

PR #16, opened 2022-02-16
Master

ritwiksharma049:
Addition of NMMA installation file.


ritwiksharma049:
Actually I edited the master directly since it was only a markdown file.
But I'll keep that in mind.

Will make changes in
https://github.com/nuclear-multimessenger-astronomy/nmma/blob/main/doc/installation.md

PR #33, opened 2022-05-20
prepping to run simulations on expanse. contains...

niharika-sravan:
minor changes to ensure interrows() variable changes are handed to the DataFrame sim to then have it be written to file such that ToO photometry and upper limits are reported with the lc in --outdir as too.csv


mcoughlin (approved):
@niharika-sravan Can we remove the outdir/0/injection_Bu2019lm.log file? I don't think we want those hanging around. Good to merge once thats in.

niharika-sravan:
Sorry that was sloppy.
--
Niharika Sravan

PR #36, opened 2022-05-24
folder with patches for sampling on EOSs

ninakuni:


ninakuni:
These are peter's patches to make bilby and pbilby work with EOSs meaning that you can sample over EOSs in parameter estimation runs. The patches were used to get the posteriors (https://github.com/nuclear-multimessenger-astronomy/nmma/blob/main/example_files/zenodo/gw_posteriors.txt) for the GW systematics study.

mcoughlin:
@tsunhopang What is the future for these patches in your mind? Will they ever be made by a PR to bilby / pbilby? We obviously will go out of date quickly like this...

PR #39, opened 2022-06-08
pymultinest instructions changed

ritwiksharma049:


doc/installation.md
```diff
@@ -98,8 +98,9 @@ Install parallel-bilby:
 
 Install pymultinest 
 
+For using the PyMultinest library, it is required to get and compile the Multinest library as well. Instructions for the same are given here:
 
-* $ conda install -c conda-forge pymultinest
```
mcoughlin, line 102: I think we should leave the conda environment as the first pass, and as a backup, installing from source. It clearly depends on the Linux distro, environment, etc.

mcoughlin (changes requested):
@ritwiksharma049 We should have both options presented.

doc/installation.md
ritwiksharma049: Made some changes. Please let me know if this is sufficient.

mcoughlin (approved):

PR #42, opened 2022-06-22
Switch to using tensorflow model __call__

dfinstad:
This patch changes light curve generation in `calc_lc` when using a tensorflow model to the model's `__call__` method instead of `predict`. The [tensorflow docs](https://www.tensorflow.org/api_docs/python/tf/keras/Model#predict) indicate the two methods are functionally equivalent, but `__call__` is more efficient when operating inside a loop with smaller batches of input data.

Profiling the code with and without this change shows a ~25x speedup on calls to `calc_lc`.


mcoughlin (approved):
@dfinstad neat trick!

PR #43, opened 2022-06-29
Update docs with an additional example using the prior

atoivonen13:
Simple change to docs showing example using an example injection file vs using a prior to create the injection file


mcoughlin (approved):
@atoivonen13 Thanks!

PR #45, opened 2022-07-02
Update of installation file

weizmannk:


doc/installation.md
```diff
@@ -96,23 +103,10 @@ Install parallel-bilby:
 
 
 
-Install pymultinest
```
mcoughlin, line 99: You can leave this part right?

mcoughlin (changes requested):
I think can leave the pymultinest stuff. Good discussion above.

weizmannk (commented):
ok, yes

weizmannk (commented):
nmma 0.0.2 requires astropy>=5.0.1, but you have astropy 4.3.1 which is incompatible.

mcoughlin:
@weizmannk Ah sorry can we finish off any place where we still pin below 5? I fixed the code so it should be fine.

weizmannk:
OK,  I install nmma on expanse.sdsc.xsede.org ... and get this error

weizmannk:
astropy =4.3 work on the old env but, the new env on jetstream2 don't work  with this version, 
I don't know what .
  File "/home/exouser/anaconda3/envs/nmm_env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 783, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'astropy>=5.0.1' distribution was not found and is required by nmma

mcoughlin:
@weizmannk What does pip install astropy==5.1 give? We could also reduce the requirement, both should still work.

weizmannk:
yes , astropy==5.1 works, 
yes the 4.3 work on my old env ... I forgot to reduce it on requirement , 
so when I install nmma, astropy=5.1 are installed , maybe this are using like dependence for another module.
so come back on 4.3 didn't work for these modules...
I will try to re install by comeback on  4.3

weizmannk:
@mcoughlin  yes it work when reduce requirement

mcoughlin:
Ok let's commit a >= 4.3 or something to resolve this.

PR #49, opened 2022-07-20
Condor script

weizmannk:
adding condor script to do observing scenarios stuff.
add request_disk
add nmma's installation  update file

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -111,6 +111,13 @@ Install pymultinest
 * $ conda install -c conda-forge pymultinest
 
 
+NOTE: In case if an error comes up during an NMMA analysis of the form:
+
+ERROR: Could not load MultiNest library "libmultinest.so" ERROR: You have to build it first, ERROR: and point the LD_LIBRARY_PATH environment variable to it!
+Then, for using the PyMultinest library, it is required to get and compile the Multinest library separately. Instructions for the same are given here:
+
+https://johannesbuchner.github.io/PyMultiNest/install.html
+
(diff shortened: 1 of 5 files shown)


doc/installation.md
```diff
@@ -88,31 +95,20 @@ OR
...
+ 
 
+OR
 
 
 * $ pip install parallel-bilby
 
 
 
-Install pymultinest
```
mcoughlin, line 99: Why did this section disappear?

mcoughlin (approved):
@weizmannk I think you lost the pymultinest instructions somehow. Once those are back, should be good to go in.

doc/installation.md
weizmannk: Do you mean that the warning error ?
Base on our last discussion https://github.com/nuclear-multimessenger-astronomy/nmma/pull/45#discussion_r912408667  

NOTE: In case if an error comes up during an NMMA analysis of the form:
ERROR:   Could not load MultiNest library "libmultinest.so"
ERROR:   You have to build it first,
ERROR:   and point the LD_LIBRARY_PATH environment variable to it!
Then, for using the PyMultinest library, it is required to get and compile the Multinest library separately. Instructions for the same are given here:
https://johannesbuchner.github.io/PyMultiNest/install.html

mcoughlin:
https://github.com/nuclear-multimessenger-astronomy/nmma/pull/49/files

Why is the PyMultinest stuff red?

mcoughlin:
https://github.com/nuclear-multimessenger-astronomy/nmma/pull/49/files

I still see red around that section. Consider copying in the file from main and re-editing.

PR #51, opened 2022-08-17
Update bilby requirement

mcoughlin:
This PR updates bilby to 1.2, which includes changes which allows the code to run as a subprocess.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 future
-bilby>=1.0.4
+bilby>=1.2.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 future
-bilby>=1.0.4
+bilby>=1.2.0
```
weizmannk, line 2: @mcoughlin  yes this works.

weizmannk (approved):
It's ok 
thanks

PR #52, opened 2022-08-24
Adding a rubin flag

weizmannk:
This PR request allows to integrate the flags of the rubin telescope while facilitating the choice if 'ztf_ToO' is true then the rubin command line is returned, and if not the analysis is done on the basis of the 'ztf' parameters.
In addition to that we add the '--train-stats' flag, if it is detected we apply it ('-- train-stats' is a flag from Ari, which gives True or False to 'ToO', which we don't use for the moment, but could be useful).


mcoughlin (approved):
LGTM!

PR #57, opened 2022-10-27
Gwemresampling docs

ninakuni:
Separate branch with all required files in example_files/tools/gwemresampling, fixed double corner images and link in documentation.


mcoughlin:
@ninakuni I do think the EOS files can live in whatever github they do, as long as they are clearly pointed to.

I think we should be moving towards a model where we have two githubs: (i) for the code and small files (this one) and (ii) for the models / big data sets mostly coming from papers.

What do you think?

mcoughlin (changes requested):
@ninakuni Very close. I suggest the prior files goes in priors/ and after that, should be good to go in.

PR #64, opened 2023-02-09
Update files to simulate detection with gwemopt 

weizmannk:
We evaluate the detection probabilities of the LSST and ZTF , on  O4 and O5 CBCs simulation , based on _gwemopt_ and the skymaps provided by the observing scenarios, in connection with the paper by  Kiendrebeogo et al.2023  .


mcoughlin (approved):
Thanks @weizmannk

PR #66, opened 2023-02-16
scipy >=1.7.1 < 1.10.0

weizmannk:
Scipy==1.10,  does not work  with bilby

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ bilby>=1.2.0
 bilby_pipe==1.0.5
 numpy>=1.9
 matplotlib>=2.0
-scipy>=1.7.1
+scipy==1.9.3
 pandas>=1.3.4
 astropy>=4.3.1
 scikit-learn==1.0.2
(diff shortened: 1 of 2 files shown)


mcoughlin:
@weizmannk Can we setup ranges rather than pinning?

weizmannk:
@mcoughlin  Yes , I will do it,

basically I think the result will be the same because look like scipy will try to install the last , 
the most recent version  agree by the code .

setup.py
```diff
@@ -77,7 +77,7 @@ def readfile(filename):
     "bilby_pipe==1.0.5",
     "numpy>=1.9",
     "matplotlib>=2.0",
-    "scipy>=1.7.1",
+    "scipy==1.9.3",
```
weizmannk, line 80:
```suggestion
    "scipy>=1.7.1, < 1.10.0",
```

requirements.txt
```diff
@@ -3,7 +3,7 @@ bilby>=1.2.0
 bilby_pipe==1.0.5
 numpy>=1.9
 matplotlib>=2.0
-scipy>=1.7.1
+scipy==1.9.3
```
weizmannk, line 6:
```suggestion
scipy>=1.7.1, < 1.10.0
```

mcoughlin (approved):
LGTM

PR #67, opened 2023-02-17
corrected command line in quick start.

ritwiksharma049:
Are SVDmodels even required for Me2017? Please check.


mcoughlin (commented):
@ritwiksharma049 I suggest adding instead some kind of note that for the non-svd models you do not need this part of the command line. For clarity.

PR #72, opened 2023-03-05
add filter check in ztf sampling

tylerbarna:
In injection.py, checks to see if current filter in loop of ztf-sampling is in the filter arguments. if not, continues to next filter without sampling. Retains requirement that filter is either g,r, or i for ztf. Shouldn't alter any other behaviour.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+
+
+
+svdmodels/Bu2019lm_lbol.pkl
+svdmodels/Bu2019lm_mag.pkl
--- a/nmma/em/injection.py
+++ b/nmma/em/injection.py
@@ -212,6 +212,8 @@ def create_light_curve_data(
         sim["mag_err"] = np.nan
 
         for filt, group in sim.groupby("passband"):
+            if filt not in args.filters.split(","):
+                continue
             data_per_filt = copy.deepcopy(data_original[filt])
             lc = interp1d(
                 data_per_filt[:, 0],
@@ -232,7 +234,6 @@ def create_light_curve_data(
 
         for filt, group in sim.groupby("passband"):
             if filt not in args.filters.split(","):
-                #print("Filter {} not in filters list".format(filt))
                 continue
             if ztf_uncertainties and filt in ["g", "r", "i"]:
                 mag_err = []
--- a/nmma/em/injection.py
+++ b/nmma/em/injection.py
@@ -29,7 +29,8 @@ def create_light_curve_data(
         lbol_ncoeff=args.injection_svd_lbol_ncoeff,
     )
 
-    bands = {1: "g", 2: "r", 3: "i"}
+    #bands = {1: "g", 2: "r", 3: "i"}
+    bands = {i+1: b for i, b in enumerate(args.filters.split(","))}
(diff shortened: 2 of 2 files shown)


tylerbarna:
To validate the functionality, should run something like 
`nmma_create_injection --injection-file example_files/sim_events/injections.dat --prior-file priors/Bu2019lm.prior --eos-file example_files/eos/ALF2.dat --binary-type BNS --n-injection 100 --original-parameters --extension json --aligned-spin
`
and then 
`    light_curve_analysis --model Bu2019lm --svd-path ./svdmodels --outdir outdir --label injection --prior priors/Bu2019lm.prior --tmin 0.1 --tmax 20 --dt 0.5 --error-budget 1 --nlive 512 --Ebv-max 0 --injection ./injection.json --injection-num 0 --injection-outfile outdir/lc.csv --generation-seed 42 --filters g --plot --remove-nondetections --ztf-uncertainties --ztf-sampling --ztf-ToO 180
`
using current version of nmma vs this PR. with current nmma version, lc.csv should ouput a lightcurve with g,r, and i despite argument. With new version, should output lightcurve with only g filter.

tylerbarna:
I added a second conditional to the ztf sampling portion of injection.py; currently, there are two instances of "for filt, group in sim.groupby("passband"):", so an additional conditional checking if the filter is in the argument list is required.

tylerbarna:
I would also recommend testing the functionality with the above code using a second model in addition to nugent-hyper, like Bu2019lm or TrPi2018

mcoughlin:
@weizmannk can you take a look?

weizmannk:
@mcoughlin  
I wait for @tylerbarna  confirmation.

To be more clear when we replace #bands = {1: "g", 2: "r", 3: "i"} by   bands = {i+1: b for i, b in enumerate(args.filters.split(","))}. 
there are always {1: "g", 2: "r", 3: "i"} in the code. 
This could generate errors in case one of the filters is not provided .

Forexample  here https://github.com/nuclear-multimessenger-astronomy/nmma/blob/b2586357f4a646e2bf0549ab4ae22fcaa7fe56e9/nmma/em/injection.py#L210

tylerbarna:
@weizmannk 
I wasn't encountering that issue, at least in the ztf sampling portion; in each of the relevant loops, it checks to see if the filter from sim['passband'] is in the arguments list; if not, it moves to the next filter

weizmannk:
@tylerbarna  ok good news, thanks

tylerbarna:
does it cause an issue when you aren't using the ztf sampling?

weizmannk (approved):
LGTM

weizmannk:
@tylerbarna  
If we give all ztf filters as input it should be ok.
So for now it's ok .

My question here was more about those who will give 1 or 2 ZTF filters instead   as arguments all 3 at the same time. 
Any way I will  do a test with the **--train-stats**  when I get free time , to confirm all thing is ok, then let you know. 

Thank you.

tylerbarna:
it seems light_curve_generation still ignores the requested filters if you call ztf-uncertainties and photometry-augmentation-times even when calling both filters and photometry-augmentation-filters

PR #73, opened 2023-03-07
Bu2022Ye model

shreyasahasram08:
Adding the changes required to create Bu2022Ye SVD models. Included in this PR are the tensorflow model as well, but those need to be updated and validated. I've not included the SVD models from Gaussian Process Regression, which are too large to commit.


mcoughlin (approved):
Looks good @shreyasahasram08

PR #76, opened 2023-03-13
pca_fit is added to the code

pouyansalehi:
Hi Michael
I think this one works.

### DIFF
--- a/nmma/em/utils.py
+++ b/nmma/em/utils.py
@@ -477,8 +477,18 @@ def calc_lc(
         # coverrors = np.dot(VA[:, :n_coeff], np.dot(np.power(np.diag(cAstd[:n_coeff]), 2), VA[:, :n_coeff].T))
         # errors = np.diag(coverrors)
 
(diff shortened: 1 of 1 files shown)


mcoughlin (changes requested):
@pouyansalehi There should be some kind of place where the default reduc_method is the old version for backwards compatibility. i.e. something like

reduc_method = svd_mag_model[filt].get('reduc_method', 'svd')

and then do the if statement with that.

pouyansalehi:
is that satisfying?

mcoughlin (approved):
@weizmannk please check when you have time.

@pouyansalehi for this to be used, this flag will also need to be added into the training. Are you also planning on opening a PR on that side?

pouyansalehi:
@mcoughlin: Yes, of course . But for now, I just wanted to prepare the code for Mouza do her PE runs ASAP.

mcoughlin:
@weizmannk Ping.

PR #77, opened 2023-03-24
remove the +1 add in eos samples 

weizmannk:


weizmannk (commented):
LGTM

mcoughlin:
@weizmannk where does this indexing issue with the EOS samples arise from?

tsunhopang:
The +1 is necessary as the casting from float to int would make the EOS drop by 1.  And why is the output file name removed?

weizmannk:
@tsunhopang  
the push was a handling error, I was trying to re-edit the code for you.  I will cancel this pull request. 
Concerning the +1, it's exactly what you said. But except that in our posteriors EOS some indices are at 5000. So the +1 is a problem. In order to solve that:

1-Either I remove the +1 in gwem_resampling and combine_EOS, 
2-Or I remove -1 from the EOS posteriors from GWdata.

Here I use the first method.

PR #82, opened 2023-04-26
removed sample_times arg in .generate_lightcurve definition

### LINKED ISSUE #81: sample times double call
When generating a lightcurve from a model, the sample times need to be provided as an argument both when the model is defined and when generate_lightcurve is called. Looking at the models in model.py, it would be sufficient to alter the code without having to provide the sample_times a second time (I imagine preserving the class parameter would touch less components outside of model.py)

tylerbarna:
changed behaviour of models doing .generate_lightcurve so they just define sample_times by retrieving it from self.sample_times. So any calls to the models are unaltered. I tried to find all calls to .generate_lightcurve and correct them, but I haven't validated all use cases.

resolves #81


mcoughlin:
@tylerbarna This is a good start. I think to have the best of both worlds, why not allow generate_lightcurves to take an optional sample_times argument? So if you pass it, great, and if not, it defaults to self.sample_times?

tylerbarna:
> @tylerbarna This is a good start. I think to have the best of both worlds, why not allow generate_lightcurves to take an optional sample_times argument? So if you pass it, great, and if not, it defaults to self.sample_times?

That's a good idea, my original thought was to have it so you only had to provide sample_times to generate_lighcurve and/or generate_spectra, but I wasn't sure if other parts of the code outside of models.py require model objects to have sample_times. There were only a handful of generate_* calls in comparison

mcoughlin:
@tylerbarna Yeah I am not unsure. I think to be safest you can for now just make it an argument that defaults to None and allows you to pass it for backwards compatability.

tylerbarna:
made the argument optional, but any code that calls generate_lightcurve without passing the sample times as a keyed argument will still break since the original version has sample_times as the first parameter

mcoughlin:
@tylerbarna For now, you can still pass samples_times=sample_times in analysis.py and other places to help mitigate behavior changing right?

tylerbarna:
yes, you should be able to, and looking through analysis.py, most calls to model objects reference the keys already

nmma/em/analysis.py
```diff
@@ -679,7 +679,7 @@ def main(args=None):
         #########################
         # Generate the lightcurve
         #########################
-        _, mag = light_curve_model.generate_lightcurve(sample_times, bestfit_params)
+        _, mag = light_curve_model.generate_lightcurve(parameters=bestfit_params)
```
mcoughlin, line 682: I guess I am thinking of here. shouldn't there be a sample_times=sample_times

nmma/em/analysis.py
```diff
@@ -693,7 +693,7 @@ def main(args=None):
 
         if len(models) > 1:
             _, mag_all = light_curve_model.generate_lightcurve(
-                sample_times, bestfit_params, return_all=True
+                prameters=bestfit_params, return_all=True
```
mcoughlin, line 696: Here too.

mcoughlin (commented):
@tylerbarna A few places where I think you can still pass: sample_times=sample_times.

Hen42rik:
The ideas of this PR are already incorporated elsewhere, therefore closing.

PR #84, opened 2023-05-10
Revise installation docs and add arm64 Mac specifics

bfhealy:
This PR revises the installation docs based on my experience installing nmma on an M1 Mac.


mcoughlin (approved):
LGTM

PR #85, opened 2023-05-10
Update wrapt_timeout_decorator import error message

bfhealy:
This PR updates a message printed if `wrapt_timeout_decorator` is not installed to make it more clear.


mcoughlin (approved):
Oops.

PR #86, opened 2023-05-11
Add more arm64 Mac installation troubleshooting

bfhealy:
This PR adds more details about installing `nmma` on an arm64 Mac based on continued experiments with the code. It also updates how code is formatted in the docs to be more distinguishable from the rest of the text.


mcoughlin (approved):
LGTM

PR #89, opened 2023-05-12
Add flag to ignore bolometric light curves in create_svdmodel.py

### LINKED ISSUE #88: Avoid error in create_svdmodel by excluding bolometric light curves
`create_svdmodel` raises an error when bolometric light curves are included in a directory that also contains filtered light curves. If we expect the names of light curve files to generally follow the format of the examples I've been using, we could avoid this error by modifying the `glob` [here](https://github.com/nuclear-multimessenger-astronomy/nmma/blob/722b02d10d95bf409da0aee1b01f4add4f32f7e1/nmma/em/create_svdmodel.py#L141) to exclude files ending with `_Lbol.dat` . This would not be useful if there is not a strict light curve file naming convention.

bfhealy:
This PR adds an `--ignore-bolometric` flag to `create_svdmodel.py` to resolve #88.

### DIFF
--- a/nmma/em/create_svdmodel.py
+++ b/nmma/em/create_svdmodel.py
@@ -121,6 +121,12 @@ def main():
         default=False,
         help="print out log likelihoods",
     )
+    parser.add_argument(
+        "--ignore-bolometric",
+        action="store_true",
+        default=False,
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #90, opened 2023-05-15
Update installation.md

tylerbarna:
added small note about installing parrallel-bilby when using WSL and solution


tylerbarna:
I know a couple of people have had issues with getting nmma installed when running `python setup.py install`, would there be an issue with adding a note about trying `pip install .` as well? That's made the install work for me on occasion.

mcoughlin (approved):
LGTM

PR #92, opened 2023-05-22
removed zero division errors

sahiljhawar:
Removed zero division, NaN, negative values and powering errors in Metzger model (em/utils.py). Light curve analysis is now approximately 2x faster.


mcoughlin:
Thanks @sahiljhawar. Please use a branch for next time.

mcoughlin (approved):
LGTM

PR #96, opened 2023-05-26
Run workflow on pull requests

bfhealy:
This PR attempts to run the workflow in `python-package.yml` on pull requests. It also standardizes the syntax of `deploy-github-pages.yml` with the other workflow files.

### DIFF
--- a/.github/workflows/build-deploy-container.yml
+++ b/.github/workflows/build-deploy-container.yml
@@ -2,7 +2,7 @@ name: CI to Docker Hub
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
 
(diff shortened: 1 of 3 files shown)


mcoughlin:
@bfhealy Maybe should take advantage of this change and also try checking 3.10?

bfhealy:
@mcoughlin That's a good idea!

bfhealy:
@mcoughlin I also noticed that the branch in `build-deploy-contained.yml` is `master` instead of `main`:
```
name: CI to Docker Hub

on:
  push:
    branches: [ master ]
```

I think that's referring to this repo, so it should probably be updated if we want this workflow running.

mcoughlin:
@bfhealy Can you update that in this PR too then?

mcoughlin (approved):
Thanks @bfhealy. LGTM.

PR #102, opened 2023-06-12
Implement wide NN to approximate GP

bfhealy:
This PR modifies the NN used by tensorflow when running `create_svdmodel`. In an effort to approximate `sklearn_gp` performance, the NN is made wider and shallower.


bfhealy:
@mcoughlin I think training and inference run decently well with the latest NN architecture in this PR. Is it ok (W.R.T. proprietary data) if I post some diagnostic plot outputs from `create_svdmodel` and `light_curve_analysis` using Bu2022Ye and the `lcs_bulla_2022` light curves?

mcoughlin:
@bfhealy yup.

bfhealy:
Since there is some strange behavior at early times, I also ran training with `tmin` set to 0.5:

injection_Bu2022Ye_lightcurves_2048_d0 6_t0 5

bfhealy:
This is the corner plot resulting from running `light_curve_analysis` on an injection using the trained model:

injection_corner

bfhealy:
And this is the (crowded) light curve output from running `light_curve_analysis`:

lightcurves

mcoughlin:
@bfhealy I suggest we roll forward with this. You can send the models to @Theodlz to have them put on Zenodo.

Thanks for making this happen.

mcoughlin (approved):
LGTM!

PR #103, opened 2023-06-12
Injections unit test

tylerbarna:
started work on injections unit test to determine why early time lightcurve values might differ between light_curve_generation and calling the lightcurveModel functions directly. While not complete, I've encountered an issue with light_curv_generation where it ignores the label argument and saves the lightcurve as 0.dat.

Beyond this issue, the next step for this unit test would be to define a function to read back in the command line lightcurve as well as a function to compare the command line and function generated lightcurves.


tylerbarna:
also, noticed there's some odd stuff with files being changed that I didn't touch being changed in the commit; will need to discard those

mcoughlin:
@tylerbarna The spirit of the test looks good. Indeed, looks like something went wonky with the commit :)

mcoughlin:
@bfhealy maybe you can help @tylerbarna out here?

bfhealy:
@mcoughlin @tylerbarna Happy to help!

bfhealy:
@tylerbarna Maybe it's best to open a new draft PR with your current progress but without the spurious commits?

tylerbarna:
> @tylerbarna Maybe it's best to open a new draft PR with your current progress but without the spurious committed files?

I agree

Issue #106, opened 2023-06-14
Early time lightcurve generation disparity


tylerbarna:
While working on an NMMA adjacent project, I found that there may be a disparity in the generated EM observations at early times (values less than 1 day) depending on the method used to generate the lightcurve. Specifically, I found a disparity between the lightcurves generated when calling light_curve_generation and those generated by the best fit parameters and calling the NMMA lightcurve model functions directly in a python script. Investigating this further, I found that there may be a disparity between the two methods of lightcurve generation (command line vs script) even when using the exact same parameter values from the same injection file.

mcoughlin:
@tylerbarna Closing this assuming that unit test solves it. Feel free to reopen if not.

Issue #53, opened 2022-09-01
plot flag file naming convention


tylerbarna:
To my knowledge, the --plot flag in light_curve_analsysis currently outputs a file named lightcurves.png. To match the naming convention of the corner plots, output file name could be changed to use the label argument (ie "[--label]_lightcurves.png")

mcoughlin:
@tylerbarna Is this now ready to close?

tylerbarna:
@mcoughlin I believe so

PR #107, opened 2023-06-14
Lightcurve Injection Unit Test

tylerbarna:
Same work/intent as PR #103, now without the commit issues. Quoting from closed PR: 

started work on injections unit test to determine why early time lightcurve values might differ between light_curve_generation and calling the lightcurveModel functions directly. While not complete, I've encountered an issue with light_curv_generation where it ignores the label argument and saves the lightcurve as 0.dat.

Beyond this issue, the next step for this unit test would be to define a function to read back in the command line lightcurve as well as a function to compare the command line and function generated lightcurves.

This is a unit test intended to investigate Issue #106 further


tylerbarna:
As mentioned before, there is an adjacent issue with nmma_create_injection where the simulation_id key is not included, which is required for light_curve_generation. The unit test currently has a workaround for this that manually adds the simulation_id key/value to the injection file.

There is also an issue with light_curve_generation where the label flag is ignored, and the file created is saved as the simulation_id value for the specific injection. This may be the intended behavior of light_curve_generation, though the label flag for light_curve_analysis uses it as a filename element in associated files (excluding the lightcurve.png plot, though this is a separate issue detailed in issue #53 )

bfhealy:
@tylerbarna I added code to include `simulation_id` in injection files where that field has been missing. Might be a good idea to remove the current workaround in this PR's code after merging the latest `main` branch.

bfhealy:
Hi @tylerbarna, I've merged the latest nmma updates and addressed the `0.dat` lightcurve issue (for now). Turns out the `--label` argument was not being used for the lightcurve filename. Instead, the `simulation_id` was being used (0 in our case). Not sure if I might be breaking something else by changing the naming convention in `create_lightcurves.py`.

I also fixed some linting errors that popped up when I made the commit. I'm now able to run `python nmma/tests/injections.py` without any errors, although I know there's more functionality to be added.

tylerbarna:
> Hi @tylerbarna, I've merged the latest nmma updates and addressed the `0.dat` lightcurve issue (for now). Turns out the `--label` argument was not being used for the lightcurve filename. Instead, the `simulation_id` was being used (0 in our case). Not sure if I might be breaking something else by changing the naming convention in `create_lightcurves.py`.
> 
> I also fixed some linting errors that popped up when I made the commit. I'm now able to run `python nmma/tests/injections.py` without any errors, although I know there's more functionality to be added.

Awesome, I'll re-merge those changes and finish up the unit test. The label and simulation_id issues are the only workarounds I can think of for running this properly, but now that they're fixed, it should be possible to properly finish the unit test

bfhealy:
Sounds good!

tylerbarna:
@bfhealy latest commit adds everything needed for testing lightcurve models to my knowledge. see 9617a09 commit notes regarding potential changes

tylerbarna:
oh, going by the unit test checks,  there's something wrong lol

mcoughlin:
@tylerbarna I copied over your test and made some small edits and it passes (locally at least). Please re-open if after taking a look anything looks problematic and such.

PR #108, opened 2023-06-14
Add simulation_id to injection file if not already there

bfhealy:
This PR adds the simulation_id to injection files that lack this column. It resets the dataframe index column and renames it to `simulation_id`. This attempts to address one of the issues mentioned by @tylerbarna in #107.


mcoughlin (approved):
LGTM

PR #109, opened 2023-06-15
Zenodo Model hosting

Theodlz:
This is an attempt at using zenodo to host models. That allows us to:
- reduce the size of nmma itself, and allow users to not have to download models they don't need.
- make sure the models are always available and centralized
- allow us to do version control. For now the code will always go fetch the latest versions of the models on zenodo, but we could do a "release" per NMMA version to pin specific version if that's the desired behavior.

That's very much of a WIP PR. I tried to call the `get_model` method a little bit everywhere it looked like it should be called. 
I haven't been able to install fftw on my mac silicon even using the instructions in the readme, so I couldn't test it so far 🤒.

Suggestion: Add some tests for these multi-filter svd models. Just one or two to test this out.
If any of the reviewers has the chance/time to review this, let me know if it works!

To see which models I've uploaded on Zenodo already, please check this page:
https://zenodo.org/record/8045280


mcoughlin (approved):
Game changer for us. Thanks @Theodlz.

PR #116, opened 2023-06-19
Update utils.py: correctly identify photosphere in metzger_lc

markoris:
absolute value was only acting on the tau array, not (tau-1), meaning that the photosphere was being identified as being at the largest ejecta radius. fixed to correctly identify photosphere as being located near minimum of (tau-1) array.


mcoughlin (approved):
LGTM

PR #117, opened 2023-06-20
Zenodo web scraping: smarter headers

Theodlz:
This is an attempt to get around the 'outdated browser' issue some users encountered. Ideally, this needs to be tried by someone with that issue before merging.

PS: Wait before merging even if confirmed that this works. I'm changing the model download method to not partially write to file.


mcoughlin (approved):
LGTM!

PR #119, opened 2023-06-22
WIP - Refresh Zenodo models list

Theodlz:


bfhealy:
@Theodlz One note as I'm testing the code - while the filter-specific models for tensorflow end with ".h5", the core files still end with ".pkl". This is raising some errors when trying to download from Zenodo.

Theodlz:
Ah I didnt know that. Let me fix it.

Theodlz:
@bfhealy what about now? I reverted some changes to make it a little easier.

bfhealy:
Thanks @Theodlz - there's one more error that I think has to do with this line:
https://github.com/nuclear-multimessenger-astronomy/nmma/blob/4d0ef5473e76df8d3204dd2bfaad705062ee14aa/nmma/em/model.py#L245

The resulting error mentions a filter I had not specified in my call to `light_curve_analysis`. I think updating this line (and the other like it) to
```
for filt in filters:
```
will fix the problem and allow the analysis to run.

Theodlz:
> Thanks @Theodlz - there's one more error that I think has to do with this line:
> 
> https://github.com/nuclear-multimessenger-astronomy/nmma/blob/4d0ef5473e76df8d3204dd2bfaad705062ee14aa/nmma/em/model.py#L245
> 
> The resulting error mentions a filter I had not specified in my call to `light_curve_analysis`. I think updating this line (and the other like it) to
> 
> ```
> for filt in filters:
> ```
> 
> will fix the problem and allow the analysis to run.

Ah yes, I had updated some of those in my first PR, might have missed a few. Let me change that now so you can try again.

Theodlz:
Done @bfhealy

bfhealy:
@Theodlz Nice, `light_curve_analysis` is running for me!

mcoughlin (approved):
Thanks for making this happen Theo.

PR #120, opened 2023-06-23
created .gitattributes for end of line definition

### LINKED ISSUE #118: End of Line
Do we use LF or CRLF for nmma? I know Unix traditionally uses LF and Windows uses CRLF, but I'm not sure what's the standard for MacOS, since most people working on the project use MacOS

tylerbarna:
Resolves #118 

created a .gitattributes file and definitions for most common file formats in the repository. This file will associate a specific line ending with all file types included in .gitattributes and default to the lf eol if not specified. This will increase compatibility between operating systems when making commits.

This changes a handful of files that were previously used CRLF line endings


tylerbarna:
The only significant file type I didn't define in the .gitattributes was the .prior file extension in case that might cause issues. However, they seem to already have lf line endings.

mcoughlin (approved):
LGTM

PR #124, opened 2023-06-23
Avoid training error for new models by specifying filters

### LINKED ISSUE #123: Error when training new model because it is not yet on Zenodo
When training a new model that is not yet on Zenodo, training finishes and the model is saved locally. But before making the final plot, the `nmma.utils.models.get_model` function is called and states that it will try local files. It then returns an empty list: https://github.com/nuclear-multimessenger-astronomy/nmma/blob/668dd0bb800a0c25ee1cb07c29c5d4c638ff947b/nmma/utils/models.py#L176 raising the following error, even though the model is saved locally:
```
Bu2019lm_tf not on Zenodo, trying local files
Traceback (most recent call last):
  File "/Users/bhealy/miniforge3/envs/nmma_env/bin/create_svdmodel", line 33, in <module>
    sys.exit(load_entry_point('nmma==0.0.8', 'console_scripts', 'create_svdmodel')())
  File "/Users/bhealy/nmma/nmma/em/create_svdmodel.py", line 223, in main
    light_curve_model = SVDLightCurveModel(
  File "/Users/bhealy/nmma/nmma/em/model.py", line 244, in __init__
    _, model_filters = get_model(
ValueError: not enough values to unpack (expected 2, got 0)
```

bfhealy:
This PR allows filter specification in `create_svdmodel` to avoid a downstream call to `get_model` that raises an error if the model is not on Zenodo (which will often be the case if training a new model). Resolves #123.

### DIFF
--- a/nmma/em/create_svdmodel.py
+++ b/nmma/em/create_svdmodel.py
@@ -227,6 +227,7 @@ def main():
         mag_ncoeff=args.svd_ncoeff,
         interpolation_type=args.interpolation_type,
         model_parameters=training_model.model_parameters,
+        filters=filts,
     )
     if args.plot:
         # we can plot an example where we compare the model performance
--- a/nmma/em/model.py
+++ b/nmma/em/model.py
@@ -181,10 +181,10 @@ class SVDLightCurveModel(object):
             self.svd_path = svd_path
 
         if self.interpolation_type == "sklearn_gp":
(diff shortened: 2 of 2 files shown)


mcoughlin:
@bfhealy I added something similar called local_only which I set to True in create_svdmodel. Hopefully this is now resolved.

PR #127, opened 2023-06-26
Model utils: download script

Theodlz:
@mcoughlin I would suggest first trying to use it as you would normally do (not on cluster, with the download occurring as the code in nmma/em tries to load the models, and then the script to make sure it all still works. I added an example of how to use the script as a comment at the end of the nmma/utils/models.py file


mcoughlin (approved):
Can you please add an entry point in the setup.py file. But it works great.

Theodlz:
Sorry havent seen your message in time @mcoughlin!

mcoughlin:
@ninakuni Please give this a try tomorrow on the cluster :)

PR #129, opened 2023-06-27
Update grb

ninakuni:
this line needs to be removed in order to make grb jet types working that do not include thetaWing parameter


mcoughlin:
@ninakuni Should this just just be like;

if "thetaWing" in new_parameters:
    grb_param_dict["thetaWing"] = new_parameters["thetaWing"]

?

mcoughlin:
Closing as we fixed this a different way.

PR #130, opened 2023-06-27
Zenodo models: try getting models from disk if Zenodo can't be accessed 

Theodlz:


Theodlz:
I tried:
- no models.yaml and zenodo access
- no models.yaml and no zenodo access
- models.yaml and no zenodo access

With this, it should look for files on disk instead of throwing an error when zenodo isn't accessible. If it can't be accessed, it will try to look for the models on disk instead and create a list out of that.

mcoughlin (approved):
LGTM. Thanks @Theodlz. @bking-astro can see if this one helps?

PR #132, opened 2023-06-27
Models list: Always add local models on top of Zenodo ones

Theodlz:


Theodlz:
@mcoughlin let's see how the tests go with this

mcoughlin (approved):
@Theodlz please merge on green

PR #133, opened 2023-06-27
Distance to redshift with astropy cosmo

mbulla:
Distance to redshift updated to use function from astropy cosmo


mcoughlin (approved):
LGTM

PR #140, opened 2023-06-28
Name generated lightcurves after label keyword

### LINKED ISSUE #137: create_lightcurves label parameter behaviour
currently, create_lightcurves uses the simulation_id value for file names rather than the label parameter, which is different from the behaviour of the label parameter in other functions. Brian fixed this in PR #107, but this was not merged

bfhealy:
This PR names generated lightcurves using the `--label` keyword, e.g. "injection_Bu2022Ye.dat". If there is more than one injection, light curves will also be subscripted with the injection's `simulation_id`, e.g. "injection_Bu2022Ye_0.dat". Resolves #137

### DIFF
--- a/nmma/em/create_lightcurves.py
+++ b/nmma/em/create_lightcurves.py
@@ -202,7 +202,9 @@ def main(args=None):
     except AttributeError:
         pass
     if refresh:
-        refresh_models_list(models_home=args.svd_path if args.svd_path not in [None, ''] else None)
+        refresh_models_list(
+            models_home=args.svd_path if args.svd_path not in [None, ""] else None
+        )
 
     seed = args.generation_seed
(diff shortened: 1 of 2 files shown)


tylerbarna (dismissed):
applies change that was lost in the unmerged #107

tylerbarna:
the unit test failure seems like a timeout unless I'm misreading it

tylerbarna:
oh, the injections unit test needs to be updated to reflect this

tylerbarna:
I think that should fix it, but could be wrong

bfhealy:
Thanks @tylerbarna! Hopefully the tests will pass now.

tylerbarna (approved):
applies change that was lost in the unmerged #107 

fixed unit test

PR #144, opened 2023-07-03
check output directory length

sahiljhawar:
To check the length of output directory (should not be longer than 64 characters.)


mcoughlin (commented):
You should also check the sampler is multinest.

sahiljhawar:
Added the condition to see if sampler is `pymultinest` or other

mcoughlin (approved):
LGTM

PR #150, opened 2023-07-17
Add time unit argument for conversions to days

bfhealy:
This PR adds a `data-time-unit` argument which can be specified in `create_svdmodel` to enable unit conversions from minutes/seconds to days (our default time unit).


mcoughlin (approved):
LGTM

PR #152, opened 2023-07-17
Remove input-shaped hidden layer from NN

### LINKED ISSUE #151: Remove additional NN layer between input and wide layer?
Between our NN's input layer and the wide (2048 neurons) layer, there is currently another layer that has the same shape as the input layer. I previously missed that this was an additional hidden layer, or I would have removed it:

https://github.com/nuclear-multimessenger-astronomy/nmma/blob/03d51ac48a80d69f54e72014a183e66e2c6a175b/nmma/em/training.py#L390

Training the same model using this updated architecture versus the current one seems to offer slight performance improvements, especially at early times; see the attached collapsar model runs comparing the current (top) and new (bottom) NN architectures. I can't make an argument for keeping this additional hidden layer (except the need to re-train any tf models and propagate this change through to existing results). 

Current:
injection_AnBa2022_lightcurves_oldNN

New:
injection_AnBa2022_lightcurves_newNN

bfhealy:
This PR removes an input-shaped hidden layer from the NN (resolves #151).

### DIFF
--- a/nmma/em/training.py
+++ b/nmma/em/training.py
@@ -387,16 +387,15 @@ class SVDTrainingModel(object):
 
             np.random.RandomState(42)
             model = Sequential()
+            # One/few layers of wide NN approximate GP
             model.add(
                 Dense(
-                    train_X.shape[1],
+                    2048,
                     activation="relu",
                     kernel_initializer="he_normal",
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #155, opened 2023-07-18
Fix bugs in model benchmarking

bfhealy:
This PR fixes an ImportError and updated model parameters for Ka2017 to enable successful benchmarking.


mcoughlin (approved):
LGTM

PR #147, opened 2023-07-14
saving bestfit to json

sahiljhawar:
Saves the best fit parameters and light curve to JSON file. Use `--bestfit` flag to save.
Issue: https://github.com/nuclear-multimessenger-astronomy/nmma/issues/138

Issue #138, opened 2023-06-28
Best Fit lightcurve results


tylerbarna:
It would be handy to have the result.json file that's produced from lightcurve analysis to include a few extra things, namely:

- Nested key for the best fit lightcurve parameters: would include the parameter keys along with their best fit values, the best fit log likelihood, log bayes factor, the index of the best parameters in the sampling array, and anything else that might be useful.
- Another key that contains the values of the best fit lightcurve in magnitudes. This would make it easier to compare the best fit against the data it's fitting, as you currently need to call the model lightcurve object again in order to get the best fit lightcurve. Should also probably contain the associated measurement times. May be worth discussing what sampling times we would want for this: should we use the same sample times as the data that's being fit, or should we use an evenly sampled time space? The former would theoretically be better for calculating residuals and whatnot, but I can see why the latter might be preferable.

sahiljhawar:
Sampling time is same as time used for light curve analysis

tylerbarna:
resolved by #147

PR #160, opened 2023-07-21
add labels to bestfit json and lightcurve plot

tylerbarna:
adds label to filename of the bestfit json, also adds a gitignore so personal vscode settings are not committed to the repo

Slightly alters work from #147, related to #138

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ __pycache__/
 nmma.egg-info/
 .DS_Store
 
+# development
+.vscode/
+
 # models
-svdmodels/Bu2019lm_lbol.pkl
-svdmodels/Bu2019lm_mag.pkl
+svdmodels/


.gitignore
```diff
@@ -5,6 +5,9 @@ __pycache__/
 nmma.egg-info/
 .DS_Store
 
+# development
+.vscode/
+
 # models
 svdmodels/Bu2019lm_lbol.pkl
```
mcoughlin, line 12: Can we modify the ignore to ignore everything in this folder now?

mcoughlin (approved):
Minor comment on the ignore file.

.gitignore
tylerbarna: the change to gitignore should do that now, no?

mcoughlin: Ah I meant in svdmodels. I noticed that it still calls out individual files

tylerbarna: ah okay, I'll open that as a separate pr then

mcoughlin: Cool

PR #162, opened 2023-07-21
consolidate lightcurve generation code

### LINKED ISSUE #161: Reduce redundancy of calling lightcurve generation in edge case
**Feature Summary**
the code used to generate lightcurves in analysis.py for plotting and the bestfit flag are redundant, in the case where both are called, the code is run twice unnecessarily

**Usage / behavior**
No change to behaviour for users 

**Alternative Solutions**
Not implementing it 

**Implementation details**
Move lightcurve generation definitions to a separate conditional above

tylerbarna:
consolidate the lightcurve generation code rather than calling it twice if bestfit and plot are called in one analysis

resolves #161

### DIFF
--- a/nmma/em/analysis.py
+++ b/nmma/em/analysis.py
@@ -673,44 +673,7 @@ def main(args=None):
     else:
         result.plot_corner()
 
-    if args.bestfit:
-        posterior_file = os.path.join(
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #163, opened 2023-07-21
Update .gitignore

tylerbarna:
now ignores all items in the svdmodels and .vscode folders. specific svdmodels files being ignored is less necessary with the zenodo changes that have been implemented recently


mcoughlin (approved):
LGTM

PR #164, opened 2023-07-21
add label param to the bestfit json and lightcurve png 

tylerbarna:
added label param to bestfit json file and lightcurve so they're not overwritten if one is running multiple fits in the same root directory


tylerbarna:
Accidentally overwrote the branch/closed #160, hence the new pr

mcoughlin (approved):
LGTM

PR #167, opened 2023-07-24
Rename 2massk filter to 2massks in AT2017gfo light curve

bfhealy:
This PR renames `2massk` to 2massks` to make the AT2017gfo light curve filter names match updated models.


mcoughlin (approved):
LGTM!

PR #168, opened 2023-07-24
Update ZTF candidate data with new filter names

bfhealy:
Like #167, this PR updates example ZTF candidate photometry, changing `g`, `r` and `i` to `ztfg`, `ztfr`, and `ztfi`.


mcoughlin (approved):
LGTM

PR #170, opened 2023-07-26
support for Python 3.11

sahiljhawar:
Support for Python 3.11 (if test passes)

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10']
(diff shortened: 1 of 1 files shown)


sahiljhawar:
@mcoughlin Test failed even for 3.8! For 3.11, the error seems to be similar what @bking-astro had, with the `lnew` in `SafeLoglikelihood()`

mcoughlin:
@sahiljhawar I assume the test must be a bit flaky for 3.8, clearly tests pass on main. For 3.11, it would be good to report a minimal failing example to https://github.com/JohannesBuchner/PyMultiNest. Can you do so?

sahiljhawar:
@mcoughlin Okay. I'll get the relevant error messages from the test and raise issue on PMN. However it's fun to see that it's backward compatible but not forward 😬

PR #172, opened 2023-07-26
Exclude nan in injection test

haukekoehn:
Sometimes the injection parameters cause the lightcurve to generate with nans. These need to be excluded in the comparison between function lightcurve and command line lightcurve


mcoughlin:
@haukekoehn personally, it would be better if the nan commit was on its own branch, and the other commits on another branch. The nan lines look fine.

tsunhopang:
I agree with @mcoughlin that I would prefer to have this changes to the tests on its own branch, without the acceleration to the GW analysis.

PR #174, opened 2023-07-26
Add --data-time-unit argument to svdmodel_benchmark

bfhealy:
This PR adds the `--data-time-unit` argument to `svdmodel_benchmark` to support grids with time units other than days. It also adds the `uvot__uvw2` filter to the column list that is used when reading grids in `.h5` format.


mcoughlin (approved):
LGTM

PR #175, opened 2023-07-27
remove nans from injection test

haukekoehn:
Sometimes, when the lightcurves generated from the injection paramete…rs contain nans. This asserts an error in the comparison np.isclose between the lightcurves from the commandline and the function, although they are the same. This is fixed by excluding the nans in the comparison now.


tsunhopang (approved):
LGTM

PR #176, opened 2023-07-27
Accelerate GW sampling

haukekoehn:
EOS are now only loaded when needed and not stored in the MultimessengerConversion object anymore. This changes the EOS2Parameters function, which is backtracked. Additionally the queue size is now always set to the size of the pool and requirements.txt are updated.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 future
 bilby>=2.1.1
 bilby_pipe>=1.1.0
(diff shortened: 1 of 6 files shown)


requirements.txt
tsunhopang, line 1: can you have these placed under production_requirement instead?

tsunhopang (approved):
LGTM

Issue #177, opened 2023-07-27
AnBa2022 parameters


bfhealy:
With recent changes to the `AnBa2022` parameters (using `log10_mtot` instead of `mtot`, etc.), there may be some remaining vestiges of the original parameters that affect benchmarking and light curve analysis. One example is how the code below takes the `log10` of the input parameters, which are now already in log-form.

https://github.com/nuclear-multimessenger-astronomy/nmma/blob/f9fce5a0cc056a1c47e035c3806aaf2ca7561b87/nmma/em/model_parameters.py#L24

When I comment the above `log10` lines, the benchmarking results for this model look much more reasonable than before.

It may also be helpful to update the `AnBa2022.prior` file to use the new log10 parameters. What do you think, @tsunhopang?

tsunhopang:
I think that's a good idea. @ninakuni could u have the prior file updated after you have the model tested on the event?

mcoughlin:
@tsunhopang @ninakuni Is this now resolved?

tsunhopang:
yes it is resolved. And now the code can handle both log and unlogged ejecta mass with the model built with log or unlogged masses.

PR #178, opened 2023-07-27
Remove log10(param) lines in model_parameters.py to reflect new AnBa2022 params

bfhealy:
This PR removes three lines that take the `log10` of parameters input to the `AnBa2022` model. The parameters in question are now input to AnBa2022 in log format, so there is no need to take another log.

See #177 for a discussion of additional steps needed to update the `AnBa2022` model across the repo.


mcoughlin (approved):
LGTM

PR #186, opened 2023-08-02
corner plot script

sahiljhawar:
Script to generate corner plots. Single posterior sample or multiple posterior samples. In case of multiple posterior datasets all posterior samples should have identical parameters.

### DIFF
--- a/tools/corner_plot.py
+++ b/tools/corner_plot.py
@@ -7,6 +7,10 @@ import os
 import seaborn as sns
 import matplotlib.patches as mpatches
 import re
+import matplotlib
+from ast import literal_eval
+
+matplotlib.use("Agg")
 
 params = {
     # latex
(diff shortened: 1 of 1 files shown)


mcoughlin:
@tsunhopang can you please review and merge if happy?

tools/corner_plot.py
```diff
@@ -0,0 +1,379 @@
+import corner
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+import argparse
+import os
+import seaborn as sns
+import matplotlib.patches as mpatches
+import re
```
tsunhopang, line 9: could u add the `matplotlib.use("agg")` so that it also work on mac

mcoughlin:
@tsunhopang Please rereview.

tsunhopang (approved):
lgtm

PR #192, opened 2023-08-06
support for coveralls

sahiljhawar:
Added support for coveralls. Not sure if the implementation is correct. Please check thouroughly.


sahiljhawar:
@mcoughlin you may have to add the `COVERALLS_REPO_TOKEN`, although not necessary

mcoughlin (approved):
LGTM, we can fix whatever we need to in future PRs

PR #195, opened 2023-08-07
Fix datatype of redshift

haukekoehn:
The convert_to_multimessenger_parameters function used to return the redshift values as astropy Quantity objects. This caused issues later, when the sample arrays were copied that contained these redshift values, turning every sample parameter to astropy Quantity objects, which becomes problematic when finally writing the posterior samples to file. 

Additionally a small bug fix induced by my previous update on how to load the EOS.


mcoughlin (approved):
LGTM

PR #198, opened 2023-08-07
Fix bug from redshift datatype conversion

haukekoehn:
the luminosity_distance_to_redshift function does not return astropy quantity objects, thus it has no value attribute. Needs to be removed from previous commit


mcoughlin (approved):
LGTM

PR #200, opened 2023-08-09
Fix API photometry filter naming and plot filename

bfhealy:
This PR removes code that used only the last character of the input filter for photometry (e.g. `r` instead of `ztfr`). It also updates the filename of the light curve plot to fix a `FileNotFoundError`.


mcoughlin (approved):
LGTM!

PR #208, opened 2023-08-25
Rename Bu2022Ye KNtimeshift prior

bfhealy:
This PR updates the name of the KNtimeshift prior for Bu2022Ye (from `trigger_time` to `KNtimeshift`).


mcoughlin (approved):
LGTM. As an aside, @bfhealy @tsunhopang at some point, we should do a PR where we do call this parameter trigger_time or similar, as KNtimeshift is clearly not representative of NMMA anymore.

Issue #101, opened 2023-06-05
Unit test for tensorflow and sklearn


mcoughlin:
We should have unit tests for both model training pathways (sklearn and tensorflow).

PR #211, opened 2023-08-28
Create dependabot.yml

sahiljhawar:
Looking at the all *_requirements.txt shows that only Sphinx is pinned to 4.4.0. And rest all required packages are above a certain version, hence the latest versions are installed. Not sure, how dependabot will be useful.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"


mcoughlin:
@sahiljhawar I would like us to figure out how to test what versions of sklearn and tensorflow break for the downloaded files, which I guess is a bit different than what dependabot does for us at the moment. I guess this should be maybe coupled with a test that commits our (smallest) sklearn / tensorflow example files and simply tries to load them in. What do you think?

sahiljhawar:
For `tensorflow` I couldn't find any installation command, excpet the docs (is it on need to know basis?). For `sklearn`, it requires >=1.0.2, hence the latest version. Did anyone faced issues with either of these libraries in the past, if that's the case, then an investigation into this will be beneficial. Also as mentioned in #101, tests already exists for both of the above mentioned libraries; however I couldn't see any output/warning/errors in the Actions, but coveralls shows the coverage.

mcoughlin:
@sahiljhawar it would still be good to find a way to specifically test the utility of the sklearn and tensorflow files we have on Zenodo.

mcoughlin:
@sahiljhawar i think we want to just have a small file locally. It's a huge burden on their servers, especially since it happens for each version of python.

sahiljhawar:
@mcoughlin We can use Git LFS to track the model files and use it in the workflow. [Found this](https://stackoverflow.com/questions/61463578/github-actions-actions-checkoutv2-lfs-true-flag-not-converting-pointers-to-act/61466160#61466160)

mcoughlin:
@sahiljhawar Can you maybe see for how long you can keep one of the downloaded files around?

PR #212, opened 2023-08-28
Changing redshift function

haukekoehn:
Because the luminosity_distance_to_redshift function of bilby doesn't scale well, a custom version of that function is added that is faster. Overall, the timesave is small, but until the luminosity_distance_to_redshift functionin bilby is not improved, we should refrain from using it.


mcoughlin:
@tsunhopang Will you have time to review this?

tsunhopang:
LGTM, waiting for the checking results

PR #215, opened 2023-09-04
multi_model_analysis command

### LINKED ISSUE #214: light_curve_analysis for multiple models
Please fill out relevant sections below; remove those that are unused.

**Feature Summary**
A function on top of the existing light_curve_analysis that allows for multiple models to be analyzed at once 

**Usage / behavior**
Call multi_model_analysis with the same arguments as light_curve_analysis, but including multiple models and priors in the arguments will result in analyses being executed for each model. Once all analyses are concluded, an optional plot flag will plot all of the model best fits together with the data.

**Alternative Solutions**
Could alter the behaviour of light_curve_analysis to have adding multiple models and priors to do what I described rather than doing a mix of the models, but I think the suggestion of another function would be better for compatibility

**Implementation details**
Additional command line function that just calls light_curve_analysis for each associated model and prior before a unique plotting feature to plot them all together onto one plot (with corner plots remaining separate)

**Additional context**
Would need to account for models taking different amounts of time to fit and also implement some type of cutoff to account for models that stall. Will also need to account for the potential of failed fits.

tylerbarna:
This is a command line function that will allow for users to provide multiple models/priors so they can all be run for one lightcurve at the same time and then directly compared against one another at the time of analysis. This is distinct from the existing functionality of light_curve_analysis and providing multiple models; the existing light_curve_analysis code allows for one to use a combination of multiple models to perform one analysis on a lightcurve (eg fitting a lightcurve to a mix of Model A and Model B, effectively Model A/B). This script would perform individual analysis of the lightcurve for each provided model and then compare the analyses against one another.

No actual functional code as of yet, but the scaffolding for a multi_model_analysis function has been created. There's a bunch of comments in the new file regarding implementation.

resolves #214

### DIFF
--- a/nmma/em/multi_model_analysis.py
+++ b/nmma/em/multi_model_analysis.py
@@ -1,0 +1,28 @@
+import argparse
+import json
+import os
+from ast import literal_eval
+
+import bilby
+import bilby.core
+import matplotlib
+import numpy as np
+import pandas as pd
+from astropy import time
+from bilby.core.likelihood import ZeroLikelihood
+
+from .analysis import get_parser
+
+matplotlib.use("agg")
+
+## TODO: make this script use all arguments from analysis.py, ideally in some dynamic way so it doesn't have to be updated every time analysis.py is updated. The only two arguments that would be different are the model and prior files, which would be based on a list of models/priors provided as an argument for this script
+## maybe also have bestfit stored as true in this script, so that it can be used to make plots of the bestfit model for each model/prior combination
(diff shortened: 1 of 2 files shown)


bfhealy (commented):
@tylerbarna This framework looks on the right track to facilitate multi-model analyses. Perhaps a good next step is to adapt your `settings.json` file in `nmma-rapid` to a config file in YAML format, as mentioned in today's meeting? Then YAML-parsing code could be added here to read all arguments.

It might be additionally helpful to add a script to generate a YAML file automatically. It could call `analysis.py get_parser` to get its arguments/defaults, ensuring that your YAML file has the latest inputs. After that, you could modify the YAML manually as desired before running `multi_model_analysis.py`.

tylerbarna:
The goal of this PR has been essentially met by #230 and other subsequent updates to the code, making this PR obsolete

PR #226, opened 2023-09-08
adds warning that KNtimeshift is deprecated

tylerbarna:
Recent update breaks compatibility with old prior files due to renaming of KNtimeshift parameter to timeshift, this PR has a few lines of code that restore compatibility but warns users that KNtimeshift will no longer work at some point in the future.

### DIFF
--- a/nmma/em/likelihood.py
+++ b/nmma/em/likelihood.py
@@ -128,7 +128,12 @@ class OpticalLightCurve(Likelihood):
             usedIdx = np.where(np.isfinite(mag_app_filt))[0]
             sample_times_used = self.sample_times[usedIdx]
             mag_app_used = mag_app_filt[usedIdx]
-            t0 = self.parameters["timeshift"]
+            try:
+                t0 = self.parameters["timeshift"]
+            except KeyError:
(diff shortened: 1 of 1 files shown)


nmma/em/likelihood.py
```diff
@@ -128,7 +128,12 @@ def log_likelihood(self):
             usedIdx = np.where(np.isfinite(mag_app_filt))[0]
             sample_times_used = self.sample_times[usedIdx]
             mag_app_used = mag_app_filt[usedIdx]
-            t0 = self.parameters["timeshift"]
+            try:
+                t0 = self.parameters["timeshift"]
+            except KeyError:
+                print("Warning: the 'KNtimeshift' parameter is deprecated as of nmma 0.0.19, please update your prior to use 'timeshift' instead")
+                t0 = self.parameters["KNtimeshift"]
+                continue
```
mcoughlin, line 136: There should not be a continue here, right?

mcoughlin (changes requested):
I think the continue should be removed.

nmma/em/likelihood.py
tylerbarna: ah yeah, not sure why I included that

tylerbarna, line 136:
```suggestion
```

mcoughlin (approved):
LGTM

PR #228, opened 2023-09-08
pinning GRB parameters

sahiljhawar:
pinning GRB parameters to avoid empty light curves error


nmma/tests/injections.py
```diff
@@ -59,6 +59,8 @@ def create_injection_from_command_line(model_name):
 
         if model_name == "nugent-hyper":
             prior_path = os.path.join( priorDir, "sncosmo-generic" + ".prior")
+        elif model_name == "TrPi2018":
+            prior_path = os.path.join( dataDir, "constTrPi2018" + ".prior") #pinning the parameter svalues in the prior file
```
mcoughlin, line 63: Can we call this TrPi2018_pinned_parameters or similar?

mcoughlin (changes requested):
This is good, should just change the parameter name.

nmma/tests/injections.py
sahiljhawar: Done

mcoughlin: @sahiljhawar TrPi2018_pinned_parameters.prior, you are missing the period.

sahiljhawar: Sorry, my bad!

mcoughlin (approved):
LGTM

PR #236, opened 2023-09-14
Set random seed for python/numpy/tf during nn training

bfhealy:
This PR sets the random seed for python/numpy/tensorflow during NN training. Previously only the numpy random state was set, leading to a variation in training results from run to run.


mcoughlin:
@bfhealy Maybe we modify this to default to some value at the command line but still allow us to control it in some way?

bfhealy:
@mcoughlin I agree, the latest commit allows `--random-seed` to be input to `create_svdmodel`.

mcoughlin (approved):
LGTM

PR #240, opened 2023-09-18
Use the same model.pkl file for all interpolation types

### LINKED ISSUE #237: Consolidate code to use the same model.pkl file for all interpolation types
The main .pkl file generated when training a model (e.g. `Bu2019lm.pkl`, `Bu2019lm_tf.pkl`) is the same regardless of interpolation type. We should consolidate the code such that it saves and loads one file per model, rather than using a different filename for each interpolation type.

bfhealy:
This PR resolves #237 by saving/loading/downloading the same file (formatted as `model.pkl`) for all interpolation types.

### DIFF
--- a/nmma/em/model.py
+++ b/nmma/em/model.py
@@ -210,7 +210,12 @@ class SVDLightCurveModel(object):
 
         self.svd_path = get_models_home(svd_path)
 
-        core_model_name = model.split("_")[0]
+        # Some models have underscores. Keep those, but drop '_tf' if it exists
+        model_name_components = model.split("_")
+        if "tf" in model_name_components:
+            model_name_components.remove("tf")
+        core_model_name = "_".join(model_name_components)
+
         modelfile = os.path.join(self.svd_path, f"{core_model_name}.pkl")
 
         if self.interpolation_type == "sklearn_gp":
--- a/nmma/em/model.py
+++ b/nmma/em/model.py
@@ -210,6 +210,9 @@ class SVDLightCurveModel(object):
 
         self.svd_path = get_models_home(svd_path)
 
+        core_model_name = model.split("_")[0]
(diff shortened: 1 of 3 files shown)


sahiljhawar:
@bfhealy Does this fix requires the currently avalaible models (which are on Potsdam server) to be renamed? 

```python
ValueError: model_name Bu2019lm_sparse_tf not found in models list
```
Why the `Bu2019.pkl` is being downloaded from Zenodo, can you please look into that?

Also can you please have a look at the Unit tests and why the `training.py` test failed and fix that as needed.

```bash
To enable the following instructions: AVX2 AVX512F FMA, in other operations, 
rebuild TensorFlow with the appropriate compiler flags.
```

bfhealy:
@sahiljhawar The latest commit passes all tests. The code does not require any existing models on Zenodo to be changed, although we may want to remove any files of the form `model_tf.pkl` in the future since they are duplicates of `model.pkl`. 

Previous commits assumed that model files would have up to one underscore (if they were trained via tensorflow). This led to an error for `Bu2019lm_sparse_tf` in the tests since it has more than one underscore. The code now supports models with additional underscores (tested locally on `AnBa2022_linear` as an example). 

Finally, I needed to make a one-line change to `load_models_list` in `models.py`. The code was checking for any files within the `models_home` directory. Now that one `.pkl` file corresponds to more than one directory of filter files, I changed the code to search for directories within that path instead of files. The directory names correlate one-to-one with a trained model, so the code now behaves as intended.

sahiljhawar:
LGTM

sahiljhawar:
@bfhealy Additionally, can you please write 1-2 liner point for release notes.

bfhealy:
@sahiljhawar Sure, please see below:

- NMMA now saves/loads the same core `.pkl` file regardless of interpolation type. Note that existing core files of the form `model_tf.pkl` are identical to `model.pkl`.

Issue #241, opened 2023-09-19
setup.py call deprecated?


tylerbarna:
I noticed that running `python setup.py install` raises two warning messages at the start regarding this being deprecated:

```
(nmma_env) [tbarna@login02 nmma]$ python setup.py install
Confirmed Python version 3.10.13 >= 3.8.0
running install
/home/tbarna/.conda/envs/nmma_env/lib/python3.10/site-packages/setuptools/_distutils/cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
!!

        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
        ********************************************************************************

!!
  self.initialize_options()
/home/tbarna/.conda/envs/nmma_env/lib/python3.10/site-packages/setuptools/_distutils/cmd.py:66: EasyInstallDeprecationWarning: easy_install command is deprecated.
!!

        ********************************************************************************
        Please avoid running ``setup.py`` and ``easy_install``.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        See https://github.com/pypa/setuptools/issues/917 for details.
        ********************************************************************************

!!
  self.initialize_options()
```

Is this something we should care about?

mcoughlin:
@tylerbarna pip install -e . also works. What do you think @sahiljhawar ?

sahiljhawar:
Since `setuptools` deprecated the use of `setup.py`, the error is shown. `setuptools` still work in legacy mode but as for future proofing we should definitely move to `pyproject.toml` based packaging. Yes, `pip instal ...` works and it will continue to work locally even with the later, the benefit would be regarding the distribution, installing the dependencies. I think even Python 3.11 or 3.12 is phasing out setuptools (and distutils) more rigorously.

PR #242, opened 2023-09-19
Replace setup.py with pyproject.toml

mcoughlin:
This PR tries to replace setup.py with pyproject.toml.


pyproject.toml
```diff
@@ -0,0 +1,132 @@
+[build-system]
+requires = ["setuptools>=45", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "nmma"
+authors = [
```
sahiljhawar, line 7: Shouldn't be same as prevoius?

mcoughlin: @sahiljhawar sure. can you fix it?

sahiljhawar: Okay

sahiljhawar:
LGTM. Resolves #241

sahiljhawar:
@mcoughlin Sorry I didn't notice but all the entry points has been renamed. Should I change all of them?

mcoughlin:
@sahiljhawar I think by convention, they should be "-" rather than "_"

sahiljhawar:
yes, but this would mean that old workflow will be broken, those who are accustomed to do `light_curve_analysis`, etc

mcoughlin:
@sahiljhawar I think we should make an announcement on list.

sahiljhawar:
Okay! Docs needs to be updated as well.

tylerbarna:
@mcoughlin @sahiljhawar  once we update the docs, perhaps we should tag this as a new release? since this alters the way users interact with most of the functions

tylerbarna:
It might also be worth (if possible) having some small period of deprecation where the old names are still callable to minimize compatibility issues with any users who updated without updating their existing codes that rely on nmma

sahiljhawar:
But if we give the upgrade window, people will still be using the old names. How often do we seriosuly care about numpy or scipy deprecations?👀

mcoughlin:
I think as long as we do a bump to 0.1.0 we can justify the change.

tylerbarna:
> I think as long as we do a bump to 0.1.0 we can justify the change.

that's a good point

tylerbarna:
A bump to 0.1.0 may also be a good time to start doing release notes

mcoughlin:
@sahiljhawar Can you write some release notes and then try a 0.1.0 release? I just pushed changes to the docs.

tylerbarna:
@mcoughlin while we're at it, we may as well remove the KNTimeshift deprecation warning

sahiljhawar:
@tylerbarna Yes!

tylerbarna:
@sahiljhawar just submitted #243 to address this

PR #246, opened 2023-09-20
Add simpler install option for pyfftw on arm64 Mac

bfhealy:
When setting up new NMMA environments on arm64 Mac architecture, I have found that the command `conda install -c conda-forge pyfftw` successfully runs after the installation fails with pip. This PR updates the installation documentation to list this simple solution as the first to try for arm64 Mac users.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -37,7 +37,7 @@ refer to the developer section below.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #247, opened 2023-09-21
Switch underscore to hyphen for all NMMA args

bfhealy:
This PR standardizes arguments across NMMA by switching any underscores to hyphens. For example, `--interpolation_type` is now `--interpolation-type` for all code using that argument. Relevant documentation has also been updated. Several automatic linting adjustments are also made to the modified code.


mcoughlin (approved):
LGTM

PR #248, opened 2023-09-25
Add code to generate slurm script for lightcurve-analysis on HPC

bfhealy:
This PR adds new code (`tools/analysis_slurm.py`) to generate a slurm script (named `slurm.sub` by default) to run `lightcurve-analysis` on an HPC resource. The code parses all current arguments in `em/analysis.py` along with slurm-specific arguments in order to allow maximum customizability.

The idea is that this code could be run one time on an HPC resource before being queued (e.g. via API calls) many times for different models/sources. Thus, a few of the defaults for `analysis.py` arguments are changed. These include `--model`, `--label`, `--trigger-time`, and `--data`, which all take environment variables by default (`$MODEL`, `$LABEL`, `$TT`, `$DATA`). These arguments can be customized upon script submission by exporting the appropriate environment variables in the call to `sbatch`, e.g:
```
sbatch --export=$MODEL=Bu2019lm,$LABEL=test,$TT=59361.0,$DATA=example_files/candidate_data/ZTF21abdpqpq.dat slurm.sub
```

Also, the default prior is `priors/$MODEL.prior`.


sahiljhawar:
LGTM.

However instead of redefining `args` it can be imported from analysis.py (`get_parser()`), this way this script need not be updated explicitly

Theodlz (approved):
LGTM! Same comment about the parser, importing it would be better.
As long as this allows us to pass the data from SP and retrieve the results, I'm very happy with this :)

sahiljhawar:
LGTM. Is it merge ready?

bfhealy:
@sahiljhawar Yes, ready to merge!

PR #249, opened 2023-09-26
Allow analysis_slurm.py to handle a mix of slurm and analysis.py arguments

bfhealy:
This PR allows `analysis_slurm.py` to handle a mix of slurm-specific and `analysis.py` arguments by adjusting the way the code gets the names of arguments from `analysis.py`.


mcoughlin (approved):
LGTM

PR #252, opened 2023-09-28
Fixes for multimessenger analysis

haukekoehn:


nmma/pbilby/parser/analysis.py
```diff
@@ -12,7 +12,7 @@ def create_nmma_analysis_parser(sampler="dynesty"):
 
     analysis_parser = argparse.ArgumentParser(prog="nmma_analysis", parents=[parser])
     analysis_parser.add_argument(
-        "data-dump",
```
mcoughlin, line 15: This just means --data-dump is the argument and then the argument is args.data_dump already.

nmma/em/model.py
```diff
@@ -291,7 +291,7 @@ def __init__(
                 if filters is None:
                     self.filters = model_filters
 
-            if os.path.isfile(modelfile):
+            elif os.path.isfile(modelfile):
```
mcoughlin, line 294: This if statement is meant to stand on its own. Maybe the local_only flag is not doing what you want?

mcoughlin (changes requested):
I do not quite understand the reason for the changes in this PR yet.

nmma/pbilby/parser/analysis.py
haukekoehn: but the data-dump is a positional argument and therefore it gets passed as args.data-dump which causes trouble

mcoughlin: No I think it turns into data_dump. Can @sahiljhawar confirm the behavior perhaps?

nmma/em/model.py
haukekoehn: No, i think the local_only flag is alright, i think. This was just a quick fix from my side, where I use the Bu2022Ye model. 
I wanted to prevent it from jumping in to the else statement below, where it tries to read in the _mag_tf model, which doesn't exist for Bu2022Ye. 

As i said this is rather hacky and now that i look at it again probably inappropriate. I think the issue is that for Bu2022Ye os.path.isfile(modelfile) is False, because modelfile actually does not point to Bu2022Ye_tf.pkl but rather to Bu2022Ye.pkl which does not exist. But I don't know how to fix this consistently for the entirety of the models out there.

nmma/pbilby/parser/analysis.py
haukekoehn: well i just checked again and when i change it back to data-dump it becomes args.data-dump at least in pbilby/analysis/main.
Screenshot from 2023-09-28 20-14-35

sahiljhawar: No, it does not gets changed to `data_dump`. It remains data-dump and it cannot be accessed as an attribute of `args`

sahiljhawar: I would suggest to change data_dump from position argument (`data_dump`) to a normal argument (`--data-dump`) and see if that works.

mcoughlin: @haukekoehn @sahiljhawar Ah that is presumably the difference. Sorry, I didn't realize that it maintains its name if positional.

sahiljhawar: @haukekoehn `args.data-dump` is not a valid syntax in case the value of `data-dump` needs to be extracted it should be done as such `getattr(args, "data-dump"))`. Also, if positional arguments are being used then in CLI, then only the order matters and neither of the two works: `--data-dump`, `--data_dump`

nmma/em/model.py
sahiljhawar: @haukekoehn If `_tf.pkl` is the problem, maybe try stripping off the `_tf` from `Bu2022Ye_tf.pkl` and use then use it with `--local-only`

nmma/pbilby/parser/analysis.py
haukekoehn: yeah now i changed it to --data-dump

nmma/em/model.py
haukekoehn: i did it the other way around, where now `_tf` is added to the `modelfile` variable, when the interpolation type is tensorflow. Maybe we can chat tomorrow about this.

mcoughlin (commented):
I am mostly ok with this as long as it works. Just unsure given the move away from the _tf pickle files.

sahiljhawar:
@mcoughlin @bfhealy Maybe for now keep support for `_tf` and raise deprecation warning until the zenodo models are updated. And then release a minor version after this issue is patched?

mcoughlin:
@sahiljhawar @bfhealy @haukekoehn I agree. Can we add an if else at the _tf check instead of just always looking for _tf?

bfhealy:
@mcoughlin @sahiljhawar I agree with this as well. I can add a temporary check for `<model>_tf.pkl` if `<model>.pkl` can't be found. This may also require a temporary modification to `get_model` in `models.py`, which currently strips the `_tf` suffix off of the .pkl files it attempts to download.

sahiljhawar:
@mcoughlin Merge-ready?

mcoughlin:
@sahiljhawar good for me!

PR #253, opened 2023-10-01
add support for time dependent (and filter dependent) systematic error 

sahiljhawar:
This PR adds support for sampled systematic error and time (and filter) dependent (interpolated) systematic error. For optimum performance it is recommended that the filters to be used for systematic biases should be less than or equal to the no of filters being used for the inference. If the no of filters for systematic biases is more than the no of `--filters` then the sampler will unnecessarily sample the priors which are of no use. For eg. if the filters in `--filters` is `u,g,r,i,z,y,J,H,K` and you want to independently sample `r` and `i` filters then the yaml should be

```yaml
config:
  withTime:
    value: true
    filters: ["r", "i", null]
    ...
    ...
```
null will be a single array which will be used for all filter excluding `r` and `i`.


sahiljhawar:
@mcoughlin @tsunhopang Can you please have a look at the `ALLOWED_FILTERS` in `systematicsprior.py`?

Right now I am using Peter's svd files and filters `--filters u,g,r,i,z,y,J,H,K`

mcoughlin:
@sahiljhawar Is the issue that we don't have a list of the now standard filters somewhere?

sahiljhawar:
@mcoughlin Yes, but apart from that. I am not sure which filters should be in the allowed list. Since sdss::u and uvot::u are different u bands.

mcoughlin:
I thought we were trying to no longer allow folks to use bands without their full names?

sahiljhawar:
Okay, in that case how can filters be named as if they are variable names and not strings. Is there any standard replacements for the symbols?

mcoughlin:
Not sure I understand the question, but we do have code here: https://github.com/nuclear-multimessenger-astronomy/nmma/blob/main/nmma/em/utils.py#L153

sahiljhawar:
Basically a list of filters which is pass to `—-filters`

mcoughlin:
Maybe we just hard code it in list form somewhere then

sahiljhawar:
Seems like a good idea.

nmma/em/analysis.py
```diff
@@ -632,9 +638,11 @@ def analysis(args):
         trigger_time=trigger_time,
         tmin=args.tmin,
         tmax=args.tmax,
+        dt=args.dt,
```
tsunhopang, line 641: what is this dt used for? does it has to be the same as the sample_times' dt?

tsunhopang: if not, it would be best with another time interval

nmma/em/likelihood.py
```diff
@@ -67,20 +71,18 @@ def __init__(
...
         elif isinstance(error_budget, dict):
             for filt in self.filters:
                 if filt not in error_budget:
                     raise ValueError(f"filter {filt} missing from error_budget")
             self.error_budget = error_budget
 
-        processedData = utils.dataProcess(
-            light_curve_data, self.filters, self.trigger_time, self.tmin, self.tmax
-        )
+        processedData = utils.dataProcess(light_curve_data, self.filters, self.trigger_time, self.tmin, self.tmax)
```
tsunhopang, line 83: does this still respect the character limit?

nmma/em/likelihood.py
```diff
@@ -97,17 +99,13 @@ def __init__(
         self.verbose = verbose
 
     def __repr__(self):
-        return self.__class__.__name__ + "(light_curve_model={},\n\tfilters={}".format(
-            self.light_curve_model, self.filters
-        )
+        return self.__class__.__name__ + "(light_curve_model={},\n\tfilters={}".format(self.light_curve_model, self.filters)
```
tsunhopang, line 102: again, the character limit

nmma/em/likelihood.py
```diff
@@ -97,17 +99,13 @@ def __init__(
...
 
     def noise_log_likelihood(self):
         return 0.0
 
     def log_likelihood(self):
-        lbol, mag_abs = self.light_curve_model.generate_lightcurve(
-            self.sample_times, self.parameters
-        )
+        lbol, mag_abs = self.light_curve_model.generate_lightcurve(self.sample_times, self.parameters)
```
tsunhopang, line 108: again, the character limit

nmma/em/likelihood.py
```diff
@@ -120,9 +118,7 @@ def log_likelihood(self):
         for filt in self.filters:
             mag_abs_filt = utils.getFilteredMag(mag_abs, filt)
             if self.parameters["luminosity_distance"] > 0.0:
-                mag_app_filt = mag_abs_filt + 5.0 * np.log10(
-                    self.parameters["luminosity_distance"] * 1e6 / 10.0
-                )
+                mag_app_filt = mag_abs_filt + 5.0 * np.log10(self.parameters["luminosity_distance"] * 1e6 / 10.0)
```
tsunhopang, line 121: again, the character limit

nmma/em/likelihood.py
```diff
@@ -144,20 +140,72 @@ def log_likelihood(self):
...
+                            if yaml_filt is not None and filt in yaml_filt:
+                                if len(yaml_filt) == 1:
+                                    filters_to_use = yaml_filt[0]
+                                else:
+                                    filters_to_use = "___".join(yaml_filt)
+                                filter_group_finite_idx_match = True
+                                break
+                    if not filter_group_finite_idx_match:
+                        filters_to_use = "all"
+                    data_sigma = np.sqrt(data_sigma**2 + (globals()[f"sys_err_{filters_to_use}_interped"](data_time)) ** 2)
```
tsunhopang, line 203: why `global` here?

nmma/em/likelihood.py
```diff
@@ -186,15 +234,33 @@ def log_likelihood(self):
...
+                                    if len(yaml_filt) == 1:
+                                        upperlim_sigma = globals()[f"sys_err_{yaml_filt}_interped"](data_time)[infIdx]
+                                    else:
+                                        filters_to_use = "___".join(yaml_filt)
+                                        upperlim_sigma = globals()[f"sys_err_{filters_to_use}_interped"](data_time)[infIdx]
+                                    filter_group_infinite_idx_match = True
+                                    break
+                        if not filter_group_infinite_idx_match:
+                            filters_to_use = "all"
+                            upperlim_sigma = globals()[f"sys_err_{filters_to_use}_interped"](data_time)[infIdx]
```
tsunhopang, line 255: again, why `global` here?

nmma/em/systematicsprior.py
```diff
@@ -0,0 +1,148 @@
...
+from pathlib import Path
+
+
+class ValidationError(ValueError):
+    def __init__(self, key, message):
+        super().__init__(f"Validation error for '{key}': {message}")
+
+
+ALLOWED_FILTERS = [
+    "u",
```
tsunhopang, line 11: these are the old filter names, and this functionality should be made flexible and with a check on comparing the data filters with the systematics filters in the analysis.py

sahiljhawar: Yes, this will be resolved later, before the final merge.

nmma/em/analysis.py
sahiljhawar: sorry, this shouldn't be there. I forgot to remove this from likelihood and analysis

nmma/em/likelihood.py
sahiljhawar: I changed my local formatting length so maybe that re-formatted this as well. apart from linting. code will work fine

nmma/em/likelihood.py
sahiljhawar: Using globals in the `all` case as well just to homogenise all the dynamically allocated variable calls

nmma/em/likelihood.py
sahiljhawar: https://github.com/nuclear-multimessenger-astronomy/nmma/pull/253#discussion_r1347374777

nmma/em/likelihood.py
sahiljhawar: https://github.com/nuclear-multimessenger-astronomy/nmma/pull/253#discussion_r1347373764

nmma/em/systematicsprior.py
sahiljhawar: The validation check (for all the grouped filters, individual filters, invalid filters, null (filter independent systematics) are validated) happens at two places, first when the systematics are loaded to the prior dict, there `systematic.main(file)` is called which validates all the inputs, see  `nmma/em/systematicsprior.py`. 

Second check happens in the likelihood as well, only if withTime is True in systematics.yaml

nmma/em/likelihood.py
sahiljhawar: Instead of `globals` another way could be monkey patching, similar to what is being done for the new `from_list` method of `PriorDict`

Issue #254, opened 2023-10-02
Add log bayes factor to bestfit params file


tylerbarna:
Please fill out relevant sections below; remove those that are unused.

**Feature Summary**
adds a key and value for the log bayes factor of the bestfit

**Usage / behavior**
no alteration to behaviour of bestfit flag, but the resulting json file has an additional key value pair

**Implementation details**
would be implemented around [here in analysis.py](https://github.com/nuclear-multimessenger-astronomy/nmma/blob/32bd04e8737c3ecefb79cba7328b041a72a29ea1/nmma/em/analysis.py#L712)

PR #255, opened 2023-10-02
write bayesfactor to json

sahiljhawar:
Solves #254

### DIFF
--- a/nmma/em/analysis.py
+++ b/nmma/em/analysis.py
@@ -744,7 +744,7 @@ def analysis(args):
 
     if args.bestfit:
         bestfit_to_write = bestfit_params.copy()
-        bestfit_to_write["ln_bayes_factor"] = result.log_bayes_factor
+        bestfit_to_write["log_bayes_factor"] = result.log_bayes_factor
         bestfit_to_write["ln_bayes_factor_err"] = result.log_evidence_err
         bestfit_to_write["Best fit index"] = int(bestfit_idx)
         bestfit_to_write["Magnitudes"] = {i: mag[i].tolist() for i in mag.keys()}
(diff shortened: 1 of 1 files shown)


nmma/em/analysis.py
```diff
@@ -744,6 +744,8 @@ def analysis(args):
 
     if args.bestfit:
         bestfit_to_write = bestfit_params.copy()
+        bestfit_to_write["ln_bayes_factor"] = round(result.log_bayes_factor, 3)
```
tylerbarna, line 747: might be better to maintain the 'log' convention even if we mean natural log, as the keys for the likelihood are 'log_likelihood' (unless I'm forgetting and the likelihood is log base 10 and bayes factor is natual log)

nmma/em/analysis.py
```diff
@@ -744,6 +744,8 @@ def analysis(args):
...
+        bestfit_to_write["ln_bayes_factor_err"] = round(result.log_evidence_err, 3)
```
tylerbarna, line 748: any particular reason for rounding? We don't round the other parameters stored in the json

sahiljhawar: it was rounded upto 3 places in Bilby, so just adopting that :)

tylerbarna: I'd probably just leave it as the raw number, that way it can be rounded as needed

nmma/em/analysis.py
```diff
@@ -744,6 +744,8 @@ def analysis(args):
 
     if args.bestfit:
         bestfit_to_write = bestfit_params.copy()
+        bestfit_to_write["ln_bayes_factor"] = result.log_bayes_factor
```
tylerbarna, line 747:
```suggestion
        bestfit_to_write["log_bayes_factor"] = result.log_bayes_factor
```

nmma/em/analysis.py
```diff
@@ -744,6 +744,8 @@ def analysis(args):
...
+        bestfit_to_write["ln_bayes_factor_err"] = result.log_evidence_err
```
tylerbarna, line 748:
```suggestion
        bestfit_to_write["log_bayes_factor_err"] = result.log_evidence_err
```

tylerbarna (approved):

Issue #256, opened 2023-10-03
trigger-time flag setting


tylerbarna:
**Describe the bug**
Currently, trigger-time is an optional flag for light_curve_analysis, but it's effectively a mandatory parameter in order for analysis to run. For example, for a lightcurve with an initial time of t=4000, not setting the trigger time will cause the log likelihood to always be negative infinity, which I suspect is due to the trigger time being set to something like t=0 if not specified

**To Reproduce**
1. generate an injection and lightcurve for any model
2. run light-curve-analysis without specifying trigger-time
3. see as your terminal is filled with low likelihood warnings

**Expected behavior**
lightcurve analysis should proceed regardless of whether or not the trigger time is specified. Trigger time should default to the first data point time, with the option of specifying a different time

**Platform information:**
 - NMMA version: all versions
 - Platform: all versions 

**Additional context**
This should be fixed by 
- [x] warning users that they have not set a trigger time
- [x] automatically setting the trigger time to be equal to the time of the first data point

Alternatively, we could modify the trigger-time behaviour in the backend to be relative to the start time of the lightcurve, which is how the tmin and tmax parameters work, but that would be a discussion to have with @mcoughlin, as I'm not sure how much modification that would require

sahiljhawar:
Instead of raising warning, why not make it a required parameter?

tylerbarna:
> Instead of raising warning, why not make it a required parameter?

I think the default use case is to set the trigger time as the first data point, so I believe it would make more sense to have the trigger time set automatically to be that with the optional choice to set a different trigger time

sahiljhawar:
If I undestand it correctly, trigger time is set to the first data point if the data file is available else it defaults to the sampling tmin. In either cases a warning should be raised that trigger time not given and a default value will be used (based on previous)

tylerbarna:
@sahiljhawar I believe the current behaviour is that the trigger time is set to 0 if the trigger time parameter isn't set, my previous comment was moreso regarding what the ideal implementation would be

PR #259, opened 2023-10-06
check and set trigger time

sahiljhawar:
Solves #256, incomplete
@tylerbarna Can you have a look at this?

Do not merge!

### DIFF
--- a/nmma/em/analysis.py
+++ b/nmma/em/analysis.py
@@ -541,23 +541,19 @@ def analysis(args):
         try:
             data = loadEvent(args.data)
             
-            #load the minimum time to use if args.trigger is None
-            #TODO: this code will be moved below, once the structue of json is identified
-            
-            min_time = np.inf
-            for key, array in data.items():
-                min_time = np.minimum(min_time, np.min(array[:, 0]))
-
         except ValueError:
             with open(args.data) as f:
                 data = json.load(f)
                 for key in data.keys():
                     data[key] = np.array(data[key])
-
+    
         if args.trigger_time is None:
-            trigger_time=min_time #For now this will only work if data file is not in JSON or try block is executed successfully.
+            #load the minimum time as trigger time
+            min_time = np.inf
+            for key, array in data.items():
+                min_time = np.minimum(min_time, np.min(array[:, 0]))
+            trigger_time=min_time
             print(f"trigger_time is not provided, analysis will continue using a trigger time of {trigger_time}") 
-            
         else:
             trigger_time = args.trigger_time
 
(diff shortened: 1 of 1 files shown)


nmma/em/analysis.py
```diff
@@ -540,16 +540,27 @@ def analysis(args):
...
                 data = json.load(f)
                 for key in data.keys():
                     data[key] = np.array(data[key])
 
         if args.trigger_time is None:
-            raise ValueError("trigger_time required if using a data file.")
-
-        trigger_time = args.trigger_time
+            print("trigger_time is not provided, analysis will continue using the time of first data point") #maybe change it to warning instead of print
+            trigger_time=min_time #For now this will only work if data file is not in JSON or try block is executed successfully.
```
tylerbarna, line 559: We should probably support json lightcurves, at least in the format that nmma itself creates from injections

tylerbarna (changes requested):

nmma/em/analysis.py
tylerbarna: But yeah, this seems to be in line with what I was thinking for handling when a trigger time wasn't provided

sahiljhawar: @tylerbarna Do you have any json LC for test?

tylerbarna: > @tylerbarna Do you have any json LC for test?

@sahiljhawar [here's one](https://github.com/tylerbarna/dsmma_kn_23/blob/model-recovery/injections/lc_Me2017_00000.json)

nmma/em/analysis.py
```diff
@@ -540,16 +540,27 @@ def analysis(args):
...
             with open(args.data) as f:
                 data = json.load(f)
                 for key in data.keys():
                     data[key] = np.array(data[key])
 
         if args.trigger_time is None:
-            raise ValueError("trigger_time required if using a data file.")
-
-        trigger_time = args.trigger_time
+            print("trigger_time is not provided, analysis will continue using the time of first data point") #maybe change it to warning instead of print
```
tylerbarna, line 558:
```suggestion

```

nmma/em/analysis.py
```diff
@@ -540,16 +540,27 @@ def analysis(args):
...
+            trigger_time=min_time #For now this will only work if data file is not in JSON or try block is executed successfully.
+            print(f"trigger_time = {trigger_time}")
```
tylerbarna, line 560:
```suggestion
            print(f"trigger_time is not provided, analysis will continue using a trigger time of {trigger_time}") 
```

nmma/em/analysis.py
sahiljhawar: I will include the edits before merging

mcoughlin (approved):
LGTM

PR #262, opened 2023-10-10
Multimessenger Inference updates

haukekoehn:
Some small changes that previously obstructed multimessenger inference. 

- Use GenericCombineLightcurveModel to pass down the filters to the GRB model (previously the GRB didn't get any filters and tried to calculate the lightcurve for all 181)
- bug fixes when calling parameter conversion
- set DOI to permanent doi if DOI is not retrievable (see opened issue on github)
- update prior file for Bu2022Ye. Before the prior bounds exceeded the grid, now the bounds from https://arxiv.org/abs/2307.11080 are adopted


nmma/em/model.py
```diff
@@ -369,7 +368,7 @@ def observation_angle_conversion(self, parameters):
     def generate_lightcurve(self, sample_times, parameters):
         if self.parameter_conversion:
             new_parameters = parameters.copy()
-            new_parameters, _ = self.parameter_conversion(new_parameters, [])
```
tsunhopang, line 372: isn't the parameter_conversion for MMA takes not only the parameters dict but also the parameter keys?

nmma/joint/likelihood.py
```diff
@@ -2,7 +2,7 @@
 
 import numpy as np
 
-from ..em.model import SVDLightCurveModel, KilonovaGRBLightCurveModel
+from ..em.model import SVDLightCurveModel, KilonovaGRBLightCurveModel, GRBLightCurveModel, GenericCombineLightCurveModel
```
tsunhopang, line 5: Maybe remove the importing of `KilonovaGRBLightCurveModel`

nmma/em/model.py
haukekoehn: For my setup self.parameter_conversion points to `MultimessengerConversion.convert_to_multimessenger_parameters` in joint/conversion.py line 468 and that function takes only the parameters as dict. i get an error when i leave the second argument there

tsunhopang (approved):
LGTM

sahiljhawar:
@haukekoehn Ready to merge?

PR #266, opened 2023-11-02
Add flag to skip sampling in analysis.py

bfhealy:
This PR adds a `--skip-sampling` flag that skips bilby sampling in analysis.py. The intended purpose of this flag is to be combined with `--plot` to generate plots based on the latest checkpoint results for interrupted runs.

PR #267, opened 2023-11-06
Add --skip-sampling flag to analysis slurm script generator

bfhealy:
This PR adds the new `--skip-sampling` analysis flag (see #266) as a wildcard in `tools/analysis_slurm.py`. This allows users to pass in `"--skip-sampling"` or an empty string to the `$SKIP_SAMPLING` environment variable to achieve different `em/analysis.py` behaviors using a single slurm script.


mcoughlin (approved):
LGTM

PR #269, opened 2023-11-10
Update lightcurve_simulation tutorial notebook and bulla files

bfhealy:
This PR updates the `lightcurve_simulation` notebook to run with the latest version of nmma. The main change is the use of sncosmo filter names that are now standard across our code. The `bulla` test grid files are also updated to use the new filter names.


mcoughlin (approved):
LGTM

PR #270, opened 2023-11-14
Modify --skip-sampling behavior to successfully generate plots

bfhealy:
This PR modifies the behavior of `analysis.py` when the `--skip-sampling` flag is set. The flag now initiates a single iteration of sampling that resumes from saved checkpoint files, allowing results/plots to be generated. This fixes an issue where plots failed to generate because the `posterior_results.dat` file was not saved until the conclusion of a sampling run.


mcoughlin (approved):
LGTM

PR #271, opened 2023-11-14
Remove temporary line in utils/models.py

bfhealy:
`utils/models.py` contains a line marked for removal once all `<model>_tf.pkl` files on Zenodo are replaced by `<model>.pkl` files. Since Zenodo is now up to date, this PR removes that line.


mcoughlin (approved):
LGTM

PR #272, opened 2023-11-15
Enable continued training of existing tensorflow model

bfhealy:
This PR adds a `--continue-training` flag to `create-svdmodel` that allows an existing tensorflow model to be additionally trained. The code raises an error if the user sets `--continue-training` with any other interpolation type.


mcoughlin (approved):
LGTM

PR #273, opened 2023-11-16
Update app.py for new syntax, easier use

bfhealy:
This PR updates the local API service (separate from the `nmma-api` repository) to use the latest NMMA syntax. It also modifies the analysis endpoint name and default port to mirror `nmma-api`, allowing the same SkyPortal analysis service to interact with this API or `nmma-api` depending on the user's preference.


mcoughlin (approved):
LGTM

PR #274, opened 2023-11-16
Add tensorflow requirement

bfhealy:
This PR adds a tensorflow requirement, in light of both the increased use of the module in NMMA and the relative cross-platform ease of installing it compared to several months ago.


mcoughlin:
@tsunhopang any objections?

tsunhopang (approved):
That's perfectly fine with me

PR #277, opened 2023-11-20
Remove more code made obsolote by Zenodo model updates

bfhealy:
In the spirit of #271, this PR removes additional code that was a temporary fix for _tf.pkl model files until Zenodo was updated.


mcoughlin (approved):
LGTM

PR #278, opened 2023-11-28
Add reduced lightcurve data

haukekoehn:


tsunhopang:
@mcoughlin I am confused by the test failing, do u have any idea on that?

mcoughlin:
@tsunhopang I think their API timesout sometimes. I wouldn't worry about that.

Issue #24, opened 2022-04-18
LSST sampling fails for Me2017


ritwiksharma049:
Running the command lines for Me2017 model with Rubin Sampling does not create **outdir** and **outdir/lc.csv** are not obtained.

1. nmma_create_injection --prior-file priors/Me2017.prior --eos-file example_files/eos/ALF2.dat --binary-type BNS --n-injection 100 --original-parameters --extension json

2. light_curve_analysis --model Me2017 --svd-path ./svdmodels --outdir outdir --label injection --prior priors/Me2017.prior --tmin 0.05 --tmax 20 --dt 0.01 --error-budget 1 --nlive 512 --Ebv-max 0 --injection ./injection.json --injection-num 0 --injection-outfile outdir/lc.csv --generation-seed 42 --filters u,g,r,i,z,y --plot --remove-nondetections --rubin-ToO --rubin-ToO-type BNS --injection-detection-limit 23.9,25.0,24.7,24.0,23.3,22.1

mcoughlin:
Running these exact lines, I get in the output directory:

ls outdir
injection.log			lc.csv
injection_corner.png		lightcurves.png
injection_posterior_samples.dat	pm_injection
injection_result.json

ritwiksharma049:
why ls outdir? did you mean cd outdir?

Also, 

injection.log: command not found

ritwiksharma049:
The command lines in my case do not produce even the outdir directory. Probably something with my machine, which I am not sure of.

PR #279, opened 2023-11-29
NMMA models - multiple sources

Theodlz:
This PR makes it possible to point either to gitlab or zenodo to fetch the models.

We make gitlab the default, as this is the one we are able to actively update. Also, we make sure that a failure to use one source of models triggers a retry with another source, until we run out of options. Here we have just 2 of them, but we can imagine having multiple ones in the future.

Maybe, later on, we'll want to have the models.yaml file live in a seperate repo or NMMA itself, and have a pointer to each source for each models, to have a more robust system. Something to think about.

For context, we opted for gitlab to bypass the very strict upload and bandwidth limits imposed by github when using git lfs (a module that allows for the upload of large binary files). Here's the gitlab repo: https://gitlab.com/Theodlz/nmma-models. There, we make use of compression to reduce the individual file size. This should help with downloading speeds, as well as maintaining the repo's full size small enough.

Also added some tests to improve the coverage a little bit :)

### DIFF
--- a/doc/models.md
+++ b/doc/models.md
@@ -3,7 +3,7 @@
 
 NMMA will be under continuous development meaning that new astrophysical models such as for kilonovae, supernovae, gamma-ray bursts and so on will be implemented over time. Therefore, it is worthwhile to check out the `model_parameters_dict` in models.py from time to time.
 
-We make the interpolated models available here: https://zenodo.org/record/8164628, which are automatically downloaded upon their first request.
+We make the interpolated models available here: https://zenodo.org/record/8039909, https://gitlab.com/Theodlz/nmma-models which are automatically downloaded upon their first request.
 
 ### Kilonovae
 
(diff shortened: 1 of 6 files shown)


bfhealy (approved):
LGTM!

Theodlz:
> LGTM!

Thanks for the review Brian! Just to be sure, did that new code work fine for you (i.e. have you had a chance to try the code0?

bfhealy:
@Theodlz Yes, I tried it out by forcing the `source` to be `zenodo` and `gitlab` and it worked in both cases. It did help me identify an unrelated issue with `analysis.py` that I'll be posting about soon.

Theodlz:
> @Theodlz Yes, I tried it out by forcing the `source` to be `zenodo` and `gitlab` and it worked in both cases. It did help me identify an unrelated issue with `analysis.py` that I'll be posting about soon.

Great! Do you need that issue to be addressed before this is merged, or should I wait before merging this?

bfhealy:
I run into this issue (#24) even on main, so I think you're good to merge this even before we resolve it.

PR #282, opened 2023-12-01
Add ZTF23aaxeacr lightcurve to example_files

bfhealy:
This PR adds the lightcurve for ZTF23aaxeacr to `example_files/candidate_data`. This lightcurve contains ZTF and SDSS magnitudes.


mcoughlin:
@bfhealy Looks like we have some duplicates in the list. Maybe we can clean it a bit? Might imply we also might want to clean the inputs coming out of SP (perhaps this is alert and FP @Theodlz?).

mcoughlin (approved):
LGTM

PR #285, opened 2023-12-06
Add support for user-specified filters in API analyses

bfhealy:
This PR adds support for a list of filters to be passed to the API for analysis. The local API is modified here, along with the code to generate a slurm script for Expanse/HPC resources. Additionally, a two-line change to `analysis.py` ensures that the code will behave as expected if the string `"None"` (rather than `None`) is given to the `--filters` argument.

A companion PR to `nmma-api` (https://github.com/Theodlz/nmma-api/pull/13) adds compatibility for user-specified filters.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -92,11 +92,18 @@ def run_nmma_model(data_dict):
     tmin = analysis_parameters.get("tmin")
     tmax = analysis_parameters.get("tmax")
     dt = analysis_parameters.get("dt")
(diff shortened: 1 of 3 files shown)


bfhealy:
Note - this PR will become unnecessary if filters/instruments are handled at the SkyPortal level.

bfhealy:
This will be addressed at the SkyPortal level (see https://github.com/skyportal/skyportal/pull/4669)

PR #286, opened 2023-12-12
Versioning change

Theodlz:
An attempt to address the versioning differently, similar to how https://github.com/joblib/joblib does it for example.

### DIFF
--- a/nmma/pbilby/generation.py
+++ b/nmma/pbilby/generation.py
@@ -24,38 +24,39 @@ from parallel_bilby.utils import get_cli_args
 from .parser import (
     create_nmma_generation_parser,
     create_nmma_gw_generation_parser,
-    parse_generation_args
+    parse_generation_args,
 )
 from ..em.io import loadEvent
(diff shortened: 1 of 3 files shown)


sahiljhawar:
will the bumpver still work as usual? You can check using `bumpver update --patch --no-commit --no-tag-commit --no-push`

Theodlz:
Hi @sahiljhawar, yes I gave it a go (a different cmd tho, not committing anything but just running it to output in the stdout) and it did update the versions in both pyproject.toml and __init__.py

sahiljhawar:
okay, great!

@mcoughlin merge?

mcoughlin:
Yes sir. Just edit the title not to have WIP. @sahiljhawar

sahiljhawar:
@Theodlz any more changes to this?

Theodlz:
@sahiljhawar I don't believe so no! We're keeping 0.1.3 here for now, right?

sahiljhawar:
I think so.

PR #292, opened 2023-12-19
Add new version of Bulla model: Bu2023Ye

shreyasahasram08:
Modified model.py, model_parameters.py and Bu2023Ye.prior


bfhealy (approved):
LGTM, thanks Shreya! Looks like the parameter parsing is running as intended with this model. I'll merge this to help us run further testing in a consistent way. I'll be especially interested to see the training results for the subset of this grid where `Yewind` is fixed at the Bu2022Ye value.

PR #294, opened 2023-12-19
Add local-model-only flag to joint analyses

bfhealy:
This PR adds a `local-model-only` flag to joint analyses to avoid downloading models from gitlab/zenodo.

(Several automated linting updates to the code as well)

### DIFF
--- a/nmma/joint/likelihood.py
+++ b/nmma/joint/likelihood.py
@@ -2,17 +2,24 @@ from __future__ import division
 
 import numpy as np
 
-from ..em.model import SVDLightCurveModel, GRBLightCurveModel, GenericCombineLightCurveModel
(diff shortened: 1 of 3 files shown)


bfhealy:
@bking-astro Once merged, this should allow you to skip the model download process by adding `local-model-only=True` to the `## EM arguments` section of your `config.ini` file.

sahiljhawar:
@bking-astro Can you test this locally and confirm if this works?

bfhealy:
Update: the flag appears to be working as intended, but we're working through what may be a cluster-specific issue with mpi.

sahiljhawar:
LGTM

sahiljhawar (approved):
LGTM

PR #297, opened 2023-12-27
adding functions to model_parameters

bking-astro:
Redoing pr to add LANL morphology functions to model_parameters.py


bfhealy (approved):
LGTM

PR #298, opened 2023-12-29
Remove 'old-style' _mag model code

bfhealy:
This PR removes code in `em/model.py` that loads deprecated models of the form `..._mag` and `..._lbol`.

The removed code tends to raise an error message about these deprecated formats when a newer model cannot be found, which can be confusing since the user is typically not trying to load the older model format. This error message has been replaced with more helpful ones describing the code's failure to load a model using the `modelfile` path.


mcoughlin (approved):
LGTM!

PR #303, opened 2024-01-04
Fix svdmodel_benchmark filter specification, add unit test

bfhealy:
This PR fixes a bug that prevented filter specification for `svdmodel_benchmark`. In addition, the code is refactored to enable a new unit test.


mcoughlin (approved):
LGTM

PR #304, opened 2024-01-05
Add tool to convert skyportal lightcurves to nmma format

bfhealy:
This PR adds a tool to convert light curves downloaded from the skyportal/fritz frontend to NMMA format. Also, the Injections & observational data documentation is updated with an example lightcurve downloaded from SkyPortal.


mcoughlin:
@bfhealy in show photometry table, I get mag and magerr. I opened a quick PR for the export button (https://github.com/skyportal/skyportal/pull/4708), although honestly I kind of wonder if we should remove this button and focus on the photometry table, which seems better. What do you think?

bfhealy:
@mcoughlin Thanks for that PR - I also think the photometry table is the better place to be downloading lightcurves (I was testing only with the "Export Bold Light Curve to CSV" button).

mcoughlin:
@bfhealy https://github.com/skyportal/skyportal/pull/4709

mcoughlin (approved):
LGTM!

bfhealy:
@mcoughlin Any concern about the test failures? They look to be related to coveralls.

mcoughlin:
@bfhealy not really. can usually rerun those tests.

PR #309, opened 2024-01-22
Add example script containing tf training calls

bfhealy:
This PR adds a `tf_training_calls.sh` script containing example calls to tensorflow training on various model grids. It also slightly modifies and updates the training documentation to point to the new script.


mcoughlin (approved):
LGTM

PR #310, opened 2024-01-23
Refactor convert_skyportal_lcs.py for script installation

bfhealy:
This PR refactors `tools/convert_skyportal_lcs.py` to allow a new script (`convert-skyportal-lcs`) to be installed via `pyproject.toml`.

Also, the docs are updated and some example data is added to enable a unit test of this code.


mcoughlin (approved):
LGTM

PR #311, opened 2024-01-23
Update LANL model in tf_training_calls.sh

bfhealy:
This PR updates the model used for training in `tf_training_calls.sh` from LANL2022 to LANLTS2.


mcoughlin (approved):
LGTM

PR #314, opened 2024-01-24
changes to filters flag in LC gen. Check MPI in analysis.py

sahiljhawar:


nmma/em/create_lightcurves.py
```diff
@@ -69,8 +69,7 @@ def get_parser():
     parser.add_argument(
         "--filters",
         type=str,
-        help="A comma seperated list of filters to use (e.g. g,r,i). If none is provided, will use all the filters available",
-        default="u,g,r,i,z,y,J,H,K",
```
mcoughlin, line 73: Can we still give a reasonable default here?

nmma/em/create_lightcurves.py
```diff
@@ -203,9 +202,9 @@ def get_parser():
     parser.add_argument(
         "--ylim",
         type=str,
-        default="-12,-18",
+        default="22,16",
```
mcoughlin, line 205: So the existing default's isn't bad for absolute magnitudes. Maybe we have a flag for --apparent or something that automatically switches the plots?

mcoughlin (changes requested):
Just a few changes.

nmma/em/create_lightcurves.py
tsunhopang: This will default to all the available filters of the given model, so it should be fine. But the help text should not be removed but updated.

mcoughlin: Sounds good.

nmma/em/create_lightcurves.py
sahiljhawar: TL:DR - I am bad at git 
This was done by @weizmannk some weeks ago, due to merge conflicts the commit related to this is reverted and showing up as new changes in this PR and few others changes as well. I will manually merge the his changes into this one

sahiljhawar: looks like @weizmannk 's limits were wrong. ylim indeed should be positive, because lc is in absolute magnitude

mcoughlin (approved):
LGTM

tsunhopang (approved):
LGTM

PR #315, opened 2024-01-29
Let svdmodel-benchmark take comma-separated filters

bfhealy:
This PR standardizes filter input across NMMA by modifying `svdmodel_benchmark.py` to take comma-separated filters (instead of space-separated).


mcoughlin (approved):
LGTM

PR #319, opened 2024-02-09
Refactor unit tests to support local pytest runs

bfhealy:
This PR refactors the analysis and models unit tests to support local pytest runs. The changes are as follows:
- The analysis test now gets svdmodels from the `nmma/tests/data` directory
- `Bu2019nsbh` models are added to the test data directory (instead of `Bu2019lm` to stay below Github's maximum file size threshold)
- The models tests are split by zenodo and gitlab in case one site is temporarily down
- Model caching is removed from the CI workflow
- Docs are updated to mention local pytest runs


sahiljhawar:
LGTM

sahiljhawar (approved):

PR #320, opened 2024-02-12
minor update to documentation formatting

tylerbarna:
this updates the format for a few lines in the install page so they can be copied directly like other code snippets


bfhealy (approved):
LGTM

PR #325, opened 2024-02-26
Only run docker CI/website deploy workflows on primary repo

bfhealy:
This PR adds the condition `if: github.repository_owner == 'nuclear-multimessenger-astronomy'` to the docker CI and website deploy workflows to prevent these from fully running unless the repository owner is `nuclear-multimessenger-astronomy`. This will prevent test failures when pushing the latest changes to forked repos. Note that most of the `build` job for GitHub pages will still run on forked repos, since only the final `Deploy to GitHub Pages` step fails.

Additionally, an extraneous level of indentation has been removed in `build-deploy-container.yml`.

### DIFF
--- a/.github/workflows/build-deploy-container.yml
+++ b/.github/workflows/build-deploy-container.yml
@@ -8,6 +8,7 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #328, opened 2024-03-04
Restrict to Python < 3.12

sahiljhawar:
From Slack:

> sounds good to me. anyways we can't move to 3.12 until Tensorflow is available for 3.12
> but, will it help? because `scikit` error will appear before the python version error


mcoughlin (approved):
LGTM

PR #330, opened 2024-03-08
Adding functionality for using Bayestar fit input for inclination

tsunhopang:
The PR allows for the Bayestar output fits to be used as inclination prior for NMMA;

A small test case is showing an expected behaviour

Corner plot without the fit input
image

Corner plot with the fit input
image

The fit input;
image

Currently, only a fixed distance prior is supported and, by default, takes the maP sky location. Subsequent PRs will expand the functionality.

### DIFF
--- a/nmma/em/analysis.py
+++ b/nmma/em/analysis.py
@@ -378,12 +378,12 @@ def get_parser(**kwargs):
     )
 
     parser.add_argument(
-        "--fit-file",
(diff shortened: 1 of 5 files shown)


tsunhopang:
It seems that the ultrasat filters it breaking the pytest

sahiljhawar:
@tsunhopang I think you have to update the tests as well

PR #331, opened 2024-03-08
add changelog

sahiljhawar:
Automatically adds Github release changelogs to documentation.


mcoughlin (approved):
Great idea. LGTM

sahiljhawar:
@mcoughlin It works locally for me. Not sure if the token will work here as well.

PR #333, opened 2024-03-10
Allow user specify ra,dec for fits input

tsunhopang:
@mcoughlin Adding the missed functionality of letting user specify ra and dec in the prior file when using fits input

MAP iota prior with the gw170817.fits:
image

iota prior with ra = 3.44616, and dec = -0.408084:
image

### DIFF
--- a/nmma/em/prior.py
+++ b/nmma/em/prior.py
@@ -77,13 +77,28 @@ def inclination_prior_from_fits(priors, args):
     # check if the sky location is input
     # if not, the maximum posterior point is taken
     if 'ra' in priors and 'dec' in priors:
-        ra = priors['ra'].peak
-        dec = priors['dec'].peak
+        ra = np.rad2deg(priors['ra'].peak)
+        dec = np.rad2deg(priors['dec'].peak)
         print(f"Using the input sky location ra={ra}, dec={dec}")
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

Maybe should note somewhere the radian bit? Or specify somehow?

tsunhopang:
It is input via the prior file, from which the ra and dec are always assumed to be radian (the same as in bilby)

mcoughlin:
Great thanks @tsunhopang

tsunhopang:
@bfhealy could you check of the failing test for training?

bfhealy:
@tsunhopang Sure, looks like it has something to do with this error:

```
TypeError: Could not locate function 'mse'. Make sure custom classes are decorated with `@keras.saving.register_keras_serializable()`. Full object config: {'module': 'keras.metrics', 'class_name': 'function', 'config': 'mse', 'registered_name': 'mse'}
```

I'll look into it some more, probably related to the recent tensorflow 2.16.1 release.

bfhealy:
@tsunhopang I propose to merge #334 to avoid this error for now. #335 is a reminder to refactor the code the address this more thoroughly in the future.

PR #336, opened 2024-03-14
Modify analysis slurm script to include PRIOR variable

bfhealy:
This PR adds a `PRIOR` variable used by the API service in generated analysis slurm scripts (see also https://github.com/Theodlz/nmma-api/pull/14).


mcoughlin (approved):
LGTM

PR #340, opened 2024-03-19
Save json file with benchmark chi2 percentiles

bfhealy:
This PR updates `svdmodel-benchmark` to print/save a json file containing the 0, 25, 50, 75, and 100th percentiles of the reduced chi2 distributions for each filter.


tsunhopang (approved):
LGTM

PR #341, opened 2024-03-19
More consistent loading of lightcurve files

tylerbarna:
Moves the try except for loading lightcurve files into the loadEvent function in io.py. This is so calls to loadEvent will work even if the supplied file is a json file. Originally, this was precent in analysis.py, but there are other calls to loadEvent, and we should preserve file compatibility for lightcurve file loading between different functions in nmma.

### DIFF
--- a/nmma/em/analysis.py
+++ b/nmma/em/analysis.py
@@ -565,15 +565,8 @@ def analysis(args):
             lc.to_csv(args.injection_outfile)
 
     else:
-        # load the kilonova afterglow data
-        try:
-            data = loadEvent(args.data)
-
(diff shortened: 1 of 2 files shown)


mcoughlin (commented):
Maybe we should be checking for json in the name to try reading that way if ends like that?

tylerbarna:
> Maybe we should be checking for json in the name to try reading that way if ends like that?

that's probably a better implementation, I was just moving the existing code into the loadEvent function. will submit that change in a moment

bfhealy (approved):
LGTM!

PR #342, opened 2024-03-19
Utility to validate lightcurves are observed adequately

tylerbarna:
This checks lightcurve files to determine if they have a requisite number of observations in the user-supplied filters within a defined amount of time since the earliest observation. Default check is 3 observations for all included filters, but may be better to have default check for at least one filter that contains 3 observations total.


mcoughlin (commented):
I think this is fine, but instead of taking args, I think it should take the individual variables as I think it will be particularly useful for calling from analysis.py on an input light curve to end early if requested. What do you think?

bfhealy (commented):
I agree with Michael on the args - this will be a useful function to call both on its own and within other NMMA code. I'd also recommend adding a unit test for this script so we can ensure it continues to work as it evolves. We could potentially use some of the existing demo data in the repo to do this.

bfhealy (approved):
LGTM

PR #343, opened 2024-03-20
Restructure sncosmo.get_bandpass block in get_default_filts_lambdas

bfhealy:
Motivated by issues running on a cluster that cannot make external internet connections, this PR restructures a block that gets sncosmo filter bandpasses in `get_default_filts_lambdas`. The removed try/except masked informative errors raised on this cluster installation, so it is instead replaced by an if/else specifying the filters that must be loaded as `AggregateBandpass` rather than `Bandpass` objects. `megacampsf` filters are added to the list that previously included only `ultrasat`.


mcoughlin (approved):
LGTM

PR #348, opened 2024-03-25
Update of resampling methods

haukekoehn:
Added maximum-mass-constraint as a command, so one can sample from a joint GW+EM posterior and get an upper limit on the TOV mass.

The TOV solver here is different from the ones already present in
eos/tov.py, because here we are interested in the density profile of the
star, whereas the other only provides bulk properties.

Conversion factors are now fixed to astropy values.


haukekoehn:
you can find a test on the Potsdam cluster at `/home/aya/work/hkoehn/EOS_analysis/inference_multimessenger/perlmutter/maximum_mass_test`

tsunhopang:
No, I meant a CI test within the nmma repo

haukekoehn:
ah ok, like in nmma/tests?

tsunhopang:
Yes, you can check `nmma/tests/analysis.py` as an example

haukekoehn:
I think this should work now :crossed_fingers:

mcoughlin:
@haukekoehn I think usually we don't keep the EOS suite in NMMA directly. I think a better test would just be the same test but with the first 10 EOSs.

tsunhopang:
Indeed a full EOS set is way too big for nmma especially it is for testing only. I agree with @mcoughlin that 10 EOS should be sufficient

haukekoehn:
Sure (it was already reduced from 25k EOS originally :D) I would also add some documentation probably. 

Is there anywhere some info/guidelines how these tests should look like? Because i guess they should mainly show that the code runs and not produce any meaningful results?

tsunhopang (approved):
LGTM

PR #349, opened 2024-03-28
Remove LANL2022 model in favor of specific morphologies

bfhealy:
This PR updates the model list and documentation to remove LANL2022 in favor of the specific morphologies: LANLTS1, LANLTS2, LANLTP1, and LANLTP2. The original LANL2022 model was trained on all four morphologies and thus produced unphysical results. The trained LANL2022 models have also been removed from gitlab, and the four morphologies have been added.

Note that `LANL2022.prior` remains in place, since it can be used in sampling for any of the morphologies.

### DIFF
--- a/doc/models.md
+++ b/doc/models.md
@@ -55,7 +55,7 @@ The original light curves are available on GitHub [here](https://github.com/mbul
 
 #### Wollaeger et al. 2021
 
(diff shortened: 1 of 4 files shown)


mcoughlin (approved):
LGTM

PR #350, opened 2024-04-05
Update index.rst

bking-astro:
added internet-less cluster instructions


mcoughlin (approved):
LGTM

bking-astro:
I'll add later a few more things about sncosmo too

PR #351, opened 2024-04-05
add `rst` linter

sahiljhawar:
Since not everyone is familiar with `rst` and directives. It is beneficial to have a `rst` linter in order to prevent incorrect docs being pushed to documentation. This only runs whenever a `rst` file is changed.


mcoughlin (approved):
LGTM

PR #356, opened 2024-04-10
Add example slurm script used by nmma API

bfhealy:
This PR adds an example slurm script to be used when running NMMA using the API service (https://github.com/Theodlz/nmma-api) that connects with skyportal/fritz. Some instructions are commented in the script file.

A script like this one can also be generated with `tools/analysis_slurm.py`.


mcoughlin (approved):
LGTM

PR #358, opened 2024-04-12
Update README.md

tylerbarna:
change year in reference


mcoughlin:
Looks like InspireHEP keeps the original arxiv year:
 https://inspirehep.net/literature/2083145

tylerbarna:
> Looks like InspireHEP keeps the original arxiv year: https://inspirehep.net/literature/2083145

ah, good catch

PR #360, opened 2024-04-30
Fix NSBH disk mass fit

atoivonen13:
use BH spin instead of chi_eff for NSBH disk mass fit

### DIFF
--- a/nmma/joint/conversion.py
+++ b/nmma/joint/conversion.py
@@ -96,31 +96,35 @@ class NSBHEjectaFitting(object):
     def __init__(self):
         pass
 
-    def chieff2risco(self, chi_eff):
+    def chibh2risco(self, chi_bh):
 
-        Z1 = 1.0 + (1.0 - chi_eff ** 2) ** (1.0 / 3) * (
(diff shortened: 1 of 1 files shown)


tsunhopang:
as @atoivonen13 suggested, we would indeed need to adjust the usage of the risco, since in the `remnant_disk_mass_fitting` we are using it as the normalized one, but in `dynamic_mass_fitting` we are using it as the risco

atoivonen13:
@tsunhopang  In the end, the only difference comes down to passing `chi_eff` or `chi_bh`. Normally, I would simply change the function to take an argument of `chi` instead of `chi_eff` and leave a function description saying passing `chi_eff` gets you risco while `chi_bh` gets you the normalized spin. However, we would also then probably want to rename the function `chieff2risco` to `chi2risco`, which could be an issue for any other uses of this function. Otherwise we can simply leave a not in the description saying passing `chi_bh` gives you the normalized risco. Duplicating the function is seems unnecessary as the code is the same.

tsunhopang:
@atoivonen13 could u take a look and see why the two unit tests are failing? After these are fixed I will have this PR merged

tsunhopang (approved):
Lgtm

Issue #352, opened 2024-04-07
Check MPI when running SVD based analysis for the first time


sahiljhawar:
When running SVD based analysis for the first time using `mpi`, all the threads/cores starts downloading models, hence creating a speed bottlneck. Before starting the download we should check for the `MPI`. And if MPI is used, download should be done by only 1 process

PR #361, opened 2024-05-02
Prevent parallel model download when using MPI

sahiljhawar:
This PR attempts to prevent multiple downloads(write) to same model file while using MPI. Solves #352


mcoughlin (approved):
LGTM! Thanks @sahiljhawar.

PR #363, opened 2024-06-11
adding new rubin ToO strategy into nmma

bking-astro:
Added the updated Rubin ToO strategy into nmma/em/injection.py and nmma/em/analysis.py. Strategies were taken from the 2024 Rubin ToO workshop (https://lssttooworkshop.github.io/images/Rubin_2024_ToO_workshop_final_report.pdf). I will note here that we only use the 120s exposure limiting magnitudes. I think this is not an issue because Night 0 for the gold strategy and Night 1,2,3 for the silver strategy use 120s exposures. The gold Nights 1,2,3 use the 180s exposure but there is only a change of ~0.2 mag in the limiting magnitudes when going from 120s to 180s. The Night 0 silver strategy uses 30s exposure but the Night 0 observation should always be brighter than Nights 1,2,3. Let me know if I should incorporate different limiting mags for different nights.


nmma/em/utils.py
```diff
@@ -1740,3 +1740,20 @@ def parse_LANLfile(filename, key="band"):
     nrows = count - key_count - 3
 
     return nrows, keys_in_file
+
+
+def detection_limits(ztf-ToO, rubin-ToO, rubin-sampling):
```
sahiljhawar: `-` ? I think it should be `_`

sahiljhawar:
@bking-astro can you also change the relevant flags in tests or create new tests within the same files? see here for an example: https://github.com/nuclear-multimessenger-astronomy/nmma/blob/d5c7188e5137aee17f62801d04643bb162b75f5a/nmma/tests/analysis.py#L89

bking-astro:
I added the ignore_timeshift flag. All the other flags in the test shouldn't be affected because they have rubin_ToO = False and rubin_ToO_type=None.

nmma/em/analysis.py
```diff
@@ -386,21 +391,6 @@ def get_parser(**kwargs):
         help="Number of cos-iota nodes used in the Bayestar fits (default: 10)",
         default=10,
     )
-    parser.add_argument(
```
sahiljhawar, line 389: your PR does not contains these flags (ra, dec, dL). is it something that you want?

nmma/em/analysis.py
```diff
@@ -444,15 +454,6 @@ def analysis(args):
         )
     else:
         sample_times = np.arange(args.tmin, args.tmax + args.dt, args.dt)
-    print("Creating light curve model for inference")
```
sahiljhawar, line 447: this section of code is also not in the PR!

sahiljhawar:
@bking-astro I have added review comments in the `File changed`

sahiljhawar:
LGTM!

[Github actions are currently not working](https://www.githubstatus.com/)

mcoughlin:
@tsunhopang can you please review and merge?

tsunhopang (approved):
LGTM

Issue #362, opened 2024-05-15
Add citation support in the model classes


sahiljhawar:
Add citation support in the model classes similar to `dynesty`. See [here](https://dynesty.readthedocs.io/en/stable/crashcourse.html) for example. Relevant code: https://github.com/joshspeagle/dynesty/blob/master/py/dynesty/dynesty.py#L24

PR #364, opened 2024-06-23
add support for citations attribute

sahiljhawar:
Solves #362
Uses a Mixin base class with `citation` attribute to be inherited by other lightcurve classes.

Example usage:

```python
kn = SVDLightCurveModel(model="Bu2019lm",sample_times=[0,0.5],svd_path="path/to/svd",local_only=True,filters=["2massh"])
grb  = GRBLightCurveModel(sample_times=[0,0.5])

combined = GenericCombineLightCurveModel(models=[kn,grb],sample_times=[0,0.5])

combined.citation
>>> [{'Bu2019lm': ['https://arxiv.org/abs/2002.11355',
    'https://arxiv.org/abs/1906.04205']},
    {'TrPi2018': ['https://arxiv.org/abs/1909.11691']}]

kn.citation
>>>{'Bu2019lm': ['https://arxiv.org/abs/2002.11355',
   'https://arxiv.org/abs/1906.04205']}

grb.citation
>>>{'TrPi2018': ['https://arxiv.org/abs/1909.11691']}
``` 

Few citations are missing


nmma/em/model.py
```diff
@@ -99,12 +99,52 @@
     ],
 }
 
-
-class GenericCombineLightCurveModel(object):
+class Mixin:
```
mcoughlin, line 102: This looks good. But could we call this Lightcurve or some other useful name?

mcoughlin (approved):
I think this looks good, let's just use a better name than Mixin.

nmma/em/model.py
sahiljhawar: Mixin is a standard name for a such a base class. Any suggestions for other name?

mcoughlin: Ah sorry should have been more clear. I meant like LightcurveMixin or similar.

mcoughlin (approved):
LGTM

PR #368, opened 2024-07-10
Observing scenarios lightcurves sim

weizmannk:
This PR adds a description for generating the kilonova light curves using the injections.dat file from the observing scenarios simulation.


mcoughlin (approved):
Awesome work

PR #373, opened 2024-07-29
Fix pyproject.toml

mcdavis62:
When installing, it needs to look into the tools folder. This doesn't work with the new pyproject.toml. [probably worked with the old setup.py]


mcoughlin (approved):
LGTM

PR #381, opened 2024-08-21
bug fix and new function in NSBH ejecta fits

atoivonen13:
Fix the bug in the NSBH fits baryon mass calculation. Move the repeated baryon mass calculation code to its own function in the class. Fix typo in baryon mass spelling. Add link to paper and equation number for baryon mass formula


tsunhopang:
LGTM

PR #383, opened 2024-09-21
add support for time dependent (and filter dependent) systematic error

sahiljhawar:
Copy of #253 

This PR adds support for sampled systematic error and time (and filter) dependent (interpolated) systematic error. For optimum performance it is recommended that the filters to be used for systematic biases should be less than or equal to the no of filters being used for the inference. If the no of filters for systematic biases is more than the no of `--filters` then the sampler will unnecessarily sample the priors which are of no use. For eg. if the filters in `--filters` is `sdssu,ps1__g,ps1__r,ps1__i,ps1__z,ps1__y,2massj,2massh,2massks` and you want to independently sample `ps1__r` and `ps1__i` filters then the yaml should be

```yaml
config:
  withTime:
    value: true
    filters: ["ps1__r", "ps1__i", null]
    ...
    ...
```
null will be a single array which will be used for all other filters excluding `ps1__r` and `ps1__i`.

The option is still there to add the systematic error in the prior file, however changed the name of the prior to be `sys_err`

### DIFF
--- a/nmma/em/analysis.py
+++ b/nmma/em/analysis.py
@@ -420,6 +420,12 @@ def get_parser(**kwargs):
         action="store_true",
         default=False,
     )
+    parser.add_argument(
+        "--systematics-file",
+        metavar="PATH",
+        help="Path to systematics configuration file",
+        default=None,
+    )
 
     return parser
 
@@ -771,6 +777,7 @@ def analysis(args):
         error_budget=error_budget,
         verbose=args.verbose,
         detection_limit=args.detection_limit,
+        systematics_file=args.systematics_file
     )
 
     likelihood = OpticalLightCurve(**likelihood_kwargs)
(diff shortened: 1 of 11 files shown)


mcoughlin:
@sahiljhawar maybe you forgot a commit?

sahiljhawar:
@mcoughlin which commit?

mcoughlin:
@sahiljhawar i just notice the tests were failing due to a missing argument :)

sahiljhawar:
yeah, I just checked the tests were failing due to `outdir` I guess. That's why the new commit with autodelete `outdir` after each test.

mcoughlin (approved):
This looks great. Can you add a page to the docs on this with maybe a few sample plots @sahiljhawar ?

sahiljhawar:
Yeah sure, I can do that.

mcoughlin:
@sahiljhawar any update herE?

sahiljhawar:
@mcoughlin Waiting for @tsunhopang review

tsunhopang:
Sorry for the delay, I should have time to look inti it this weekend.

tsunhopang:
lgtm

Issue #386, opened 2024-11-02
Docs for systematics error


sahiljhawar:
Add documentation for systematics error.

PR #387, opened 2024-11-02
systematics docs

sahiljhawar:
#386

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,18 @@ When utilizing this code for a publication, kindly make a reference to the packa
       url={https://doi.org/10.1038/s41467-023-43932-6}
 }
 ```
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
This is excellent. Perhaps do you want to include an example plot from your paper?

mcoughlin (commented):
I also wonder if we should add your citation to the front README for when folks should cite your paper in particular if they turn on systematics.

sahiljhawar:
@mcoughlin We haven't use any "systematic corner plots" in the paper, instead we always used the systematics to do interpolation for the light curves which I feel is beyond the scope of the documentation (however I can explain it if someone wants to use it in their workflow).

Citation in front README sounds good to me.

PR #389, opened 2025-01-03
Direct eos sampling

Hen42rik:
This contains some major refactoring in order to allow direct eos sampling and allow an easier modularisation in future commits.


sahiljhawar:
@Hen42rik is this WIP or more commits will be added to this PR?

sahiljhawar:
@mcoughlin @tsunhopang should we discuss about this? How this is going to be added to the repo? On the main branch or the `direct_eos_sampling` branch (will not be available on PyPi)

mcoughlin:
@Hen42rik @sahiljhawar @tsunhopang Does it make sense to make thiis sampling its own executable for now, while the classes and everything of course get integrated as normal, and we can slowly converge to everything being called from one sampler code?

Hen42rik:
@mcoughlin @sahiljhawar @tsunhopang I am working on some more edits and fixes, I plan to ursurpate a longform-update next week or the week after to give an overview of suggested changes. I would suggest du discuss the further process then

Hen42rik:
@tsunhopang I have slightly modified a handful of the unit tests (mostly just replacing defunct args or names thereof), but otherwise I was careful to maintain backwards-compatibility. The errors in the last unit test runs were (somehow) due to github's caching which does not work nicely with Multinest. Rerunning the workflow without any changes to the code but after deleting the workflow cache has allowed all checks to pass

PR #407, opened 2025-08-15
add support for python 3.13

sahiljhawar:
add support for python 3.13

### DIFF
--- a/.github/workflows/continous_integration.yml
+++ b/.github/workflows/continous_integration.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade git+https://github.com/bitranox/wrapt_timeout_decorator.git
-          python -m pip install pytest pytest-cov flake8 pytest-aiohttp sqlparse freezegun PyJWT joblib coveralls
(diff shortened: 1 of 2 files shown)


sahiljhawar:
@mcoughlin Dependency installation in python 3.13 probably fails due to old NMMA pip package. When installing from source, it is working. Maybe a new release will fix this issue?

Also, Python 3.13 unit tests are failing due to coveralls not available on Python 3.13 as a result it is downloading old packages which causes a compatibility issue between coverage and coveralls. Should I add conditional in Python 3.13 Coveralls step to skip it for now, and merge this PR?

Thoughts

mcoughlin:
Maybe also link to any related issue in Coveralls?

mcoughlin (approved):
@sahiljhawar do what you need to do to get things working :)

sahiljhawar:
@mcoughlin I tested it locally, while the coverage and coveralls are being installed on Python 3.13. On github runners, for some reason it is installing an ancient version of coveralls which is causing this compatibility issue.

sahiljhawar:
@mcoughlin newer version of?

mcoughlin:
@sahiljhawar coveralls.

mcoughlin:
@sahiljhawar yeah or like a minimum version or something.

sahiljhawar:
@mcoughlin okay, sure

PR #409, opened 2025-10-30
Fix radec units

weizmannk:
This PR fixes an issue with the RA and DEC units, which should be in degrees ("deg") instead of radians ("rad").
It also fixes a typo in the parameter of HoNa2020 model and remove duplicate.


Hen42rik:
Closed upon private discussion with @weizmannk

PR #416, opened 2026-03-24
Bug fix: filters could be None

haukekoehn:
When doing an analysis, one should be able to leave the filters unspecified, so that in the end just the filters from the data are taken. This happens later, when filters_to_analyze is determined. But before that, it tries to find a detection_limit depending on filter, which will fail if filter is not specified.

This edit fixes this, although I am not sure if it's in the most elegant way imaginable.


Hen42rik:
This is addressed by now

PR #417, opened 2026-05-04
Dev

Hen42rik:
Move various upgrades to main, in particular for mpi-sampling with other samplers than dynesty and


tsunhopang (approved):
lgtm

PR #421, opened 2026-07-01
update pyproject for pypi

Hen42rik:
This removes the github download requirement for fiesta since this is now available via pip and prevents publication to pypi.org


tsunhopang (approved):
lgtm

PR #422, opened 2026-07-01
BUG: offset handling in captions broke in some cases

Hen42rik:
Minor edit for 1.0.1, failed for some fancy plotting uses


mcoughlin (approved):
LGTM

PR #425, opened 2026-07-04
restore 0.2.3 amplitude/distmod behavior for supernova model

### LINKED ISSUE #424: SupernovaLightCurveModel returns absolute magnitudes, distance modulus dropped in nmma-1.0.1
In nmma-0.2.x the distance modulus was applied in the likelihood (likelihood.py, `mag_app = mag_abs + 5log10(D_L·1e6/10)`). In nmma-1.0.1 this step moved into `combine_detector_data`, but `SupernovaLightCurveModel.gen_detector_lc`  overrides that path and returns magnitudes without re-applying it. The predicted light curve is therefore left on an absolute-magnitude scale instead of the apparent scale the
likelihood expects.

The model is systematically far brighter than the data across all bands; best-fit magnitudes are unphysical, the reduced chi-squared is very large, and the brightness-related parameters (luminosity distance, magnitude
boost, colour excess) are driven hard against their prior bounds rather
than converging to an interior posterior.

**Proposed fix**
Re-apply the distance modulus in `gen_detector_lc` for anchored templates, skipping SALT models (where the amplitude parameter already encodes the
apparent flux).

weizmannk:
[Fixes #424
](https://github.com/nuclear-multimessenger-astronomy/nmma/issues/424)
## Problem
`SupernovaLightCurveModel.gen_detector_lc` overrides the generic detector path and does not re-apply the distance modulus that `combine_detector_data` adds. The predicted magnitudes stay on an absolute scale, leaving the model far brighter than the data and railing the brightness parameters against their prior bounds.

## Fix
Re-apply `self.distmod` in `gen_detector_lc` for anchored templates. SALT models are exempt: their amplitude parameter already encodes the apparent flux, so adding the modulus would double-count the distance.

## Effect
- Before: best-fit magnitudes on an absolute scale; brightness parameters pinned to prior edges; very large reduced chi-squared.
- After: magnitudes on the apparent scale; brightness parameters converge to an interior posterior; reduced chi-squared of order unity.

## Notes
Change is confined to `SupernovaLightCurveModel`; no other model class is affected.

### DIFF
--- a/nmma/em/model.py
+++ b/nmma/em/model.py
@@ -384,8 +384,8 @@ class LightCurveModelContainer:
                 apparent_magnitude = mags + self.distmod + redshift_correction
                 # apparent_magnitude = utils.autocomplete_data(
                 #      observable_times,  observable_times[use_mask], apparent_magnitude[use_mask])
-            else: #no meaningful inter-/extrapolation possible
-                apparent_magnitude =  np.full_like(observable_times, np.inf)
+            else:  # no meaningful inter-/extrapolation possible
+                apparent_magnitude = np.full_like(observable_times, np.inf)
             lc_data[filt] = apparent_magnitude
 
         return (observable_times, lc_data)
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #430, opened 2026-07-22
Fix issue in mass source conversion

weizmannk:
Fixes a KeyError on mass_1_source/mass_2_source reported by @np699 : EoS conversion needs these fields, but they were only computed when an `snr` test was requested. Now the `gw` conversion step always runs when EoS conversion is used.

@Hen42rik  Could you please take a look? I left two `# FIXME Weizmann:` comments in the code. If everything looks correct, please feel free to remove them, and I will merge the PR.
Thanks

### DIFF
--- a/nmma/core/conversion.py
+++ b/nmma/core/conversion.py
@@ -13,8 +13,9 @@ from bilby.gw.conversion import (
     convert_to_lal_binary_black_hole_parameters,
     convert_to_lal_binary_neutron_star_parameters,
     generate_mass_parameters,
-    chirp_mass_and_mass_ratio_to_total_mass
+    chirp_mass_and_mass_ratio_to_total_mass,
 )
+
 
 def val_to_scalar(val):
     """Convert single-value quantities to scalars for easier handling"""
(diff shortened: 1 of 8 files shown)


mcoughlin (commented):
@weizmannk can you add a few tests that cover this?

weizmannk:
@mcoughlin Before the test, I have a serious concern.

Now that the EOS determines whether an injection is classified as BNS or NSBH, there is no longer a binary-type label independent of the EOS. While this is a reasonable change, I think it could significantly affect our statistics.

Many injections that were originally generated as BNS will now be reclassified as NSBH because the chosen EOS has a maximum neutron star mass of around 2 or bit more (M_\odot). However, in our injection setup, the neutron star upper mass limit is 3 (M_\odot), so those systems were intended to be treated as BNS. As a result, the BNS and NSBH populations will shift simply because of the EOS choice, rather than because the underlying injections have changed.

Is this expected? Should the injection type remain fixed based on how the injections were generated, instead of being reclassified according to the EOS? 
If yes we need to create classify flag to target them. Righ ?

The relevant code appears to be here:
https://github.com/nuclear-multimessenger-astronomy/nmma/blob/5771cfdb0541d4e9ff24320e5c13922b76a785de/nmma/core/conversion.py#L744-L766

Hen42rik:
I am not sure what exactly is your concern here. If you create an injection assuming an EoS, it is not meaningful to use a flag that says: You are still a NS at 3 solar masses, even though your EoS suggests otherwise. If, on the other hand, you want to make sure that your injection remains a BNS, you can impose a corresponding constraint. in the prior from which you create the injection. You could there add 
`lambda_1 = Constraint(maximum=0, name ="lambda_1")` which would enforce a BH primary or `minimum =1e-3` to enforce a NS primary
-> And actually, this would be good check to see whether we get these properties

weizmannk:
That's not quite my concern, actually. Let me be more precise.

1. Current default behavior (`KilonovaEjectaFitting`.**ejecta_parameter_conversion**, no `--binary-type`): for each row, 
it automatically decides BNS vs. NSBH from the EOS derived radii (`radius_1`>0 and `radius_2`>0 → BNS formula; otherwise => NSBH formula). So if you feed it a file of BNS systems and, for some row, one component turns out to be too massive for the EOS, it silently switches to the NSBH ejecta formula for that row instead of just reporting "no ejecta". That's the part I don't think is desired: a system you know is a BNS shouldn't silently become an NSBH computation just because one mass falls outside the EOS table.

Unless that fallback is intentional,  e.g. to increase the ejecta yield,  in which case I would like to understand why.

2. Separate issue: `--tests` **ejecta** / `--post-processing` **ejecta** only have any visible effect if `log10_mej_dyn` and `log10_mej_wind` are absent from the prior. If they are present, both flags get silently overridden by the prior sampled values, with no warning, but the user has no way to know that and reasonably expects the flag to do something.

What `--binary-type` does: it fixes both issues by making the choice explicit instead of automatic. You pass `--binary-type` BNS or --binary-type NSBH, and only that formula is applied to every row,  never a per-row automatic switch. Rows that don't produce finite ejecta under that formula are dropped, instead of silently falling back to the other formula.

This is still a stopgap: it doesn't touch `--tests` ejecta/`--post-processing` ejecta themselves, which rely on redraws — something that doesn't make sense anyway when masses come from an external injection file (a fixed mass can't be redrawn). I would d like to understand what behavior is actually wanted there before touching that part.

Does this clarify the concern?


@mcoughlin the test is done :  https://github.com/nuclear-multimessenger-astronomy/nmma/pull/430/changes/1887f9da7aec0e3c355124830f1cd1bd7ae99aa5

weizmannk:
@mcoughlin this can be merge now ?

PR #431, opened 2026-08-02
Fix bug lightcurve

weizmannk:
## Fix injection light-curve export: restore full grid, add real .csv output

Two fixes to how injection light curves are saved (both regressions from the em-folder modularisation refactor):

1. nmma/em/analysis.py:  non-detections were dropped instead of saved as upper limits (mag_error=inf), causing differently-shaped arrays per filter and crashing the CSV writer. Restored nmma 0.2.3's keep_infinite_data=True for the save only; fit behavior unchanged.
2. nmma/em/io.py, .csv extension was accepted but silently wrote nothing; fixed, now uses real commas (was space-delimited). Also fixed a missing space in the header.

I left # FIXME Weizmann comments in the code, I will remove them once everything is confirmed.

### DIFF
--- a/nmma/em/analysis.py
+++ b/nmma/em/analysis.py
@@ -25,15 +25,7 @@ def data_from_injection(args, filters, detection_limit):
         print(f"Loading existing injection lc from {inj_outfile}")
         full_data = io.load_em_observations(inj_outfile, format="model")
     else:
-        # FIXME Weizmann: nmma 0.2.3 saved the injection light curve with
         # keep_infinite_data=True (full time grid, non-detections marked
-        # mag_error=inf rather than dropped), so every filter shared the
-        # same length. That override was lost in a later refactor, so
-        # filters with different per-filter detection limits ended up
-        # differently trimmed before ever being written to disk. Restore
-        # keep_infinite_data=True for the save only; the fit's own `data`
-        # below is derived by filtering full_data the same way
-        # keep_infinite_data=False used to, so the fit is unaffected.
         full_data = create_light_curve_data(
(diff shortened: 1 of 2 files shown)


nmma/em/analysis.py
```diff
@@ -23,10 +23,31 @@ def data_from_injection(args, filters, detection_limit):
     inj_outfile = set_filename(args.label, args, "_lc")
     if os.path.isfile(inj_outfile):
         print(f"Loading existing injection lc from {inj_outfile}")
-        data = io.load_em_observations(inj_outfile, format="model")
+        full_data = io.load_em_observations(inj_outfile, format="model")
     else:
-        data = create_light_curve_data(injection_params, args, inj_model)
-        io.write_em_observations(inj_outfile, data, format="model")
+        # FIXME Weizmann: nmma 0.2.3 saved the injection light curve with
```
mcoughlin: I don't think we need this verbose a comment

nmma/em/io.py
```diff
@@ -129,72 +150,119 @@ def strict_read_csv(filename, args):
...
 
+
 def write_lc_to_json(injection_outfile, data):
     with open(injection_outfile, "w") as f:
         json.dump(data, f, cls=NumpyEncoder, indent=2)
 
-def write_lc_to_csv(outfile, data, format= "observations"):
+
+def write_lc_to_csv(outfile, data, format="observations"):
+    # FIXME Weizmann: despite the name, this always wrote space-delimited
```
mcoughlin: Same here

mcoughlin (approved):
Just shorten the comments. Otherwise merge on green.

PR #434, opened 2026-09-08
Restructure tests

Hen42rik:
This PR is mostly cosmetic to simplify upcoming development work. It addresses four issues with tests:

1.  It regroups tests into subdirectories according to the main project structure.
2.  It harmonises the way we handle pathes throughout the code, removing all occurences of os.path in favour of pathlib.Path
3.  It cuts data requirements for tests by allowing injections to be directly sampled from priors, with benefits for wider injections approaches.
4. It fixes some data inconsistencies when providing args not from command-line by removing the outdated bilby_pipe nonetypes.


ThibeauWouters (approved):
LGTM

mcoughlin (approved):
LGTM

PR #435, opened 2026-09-08
Extinction update

Hen42rik:
This pull request addresses multiple issues around extinction, refactoring the handling of kwargs in lightcurve models along the way:
1. The Ebv prior is no longer added automatically. Instead, we create a constant value in preprocessing if there is a dustmap in use or we check whether it is provided as a sampling parameter in the prior file.
Note that this  deliberately removes the „Ebv-max“ and „use-Ebv“ arguments because it seems bad practice to add priors outside the prior-file. Moreover, this allows more flexibility with the chosen prior.
2. Correspondingly, we only pursue extinction corrections if the prior contains such a parameter. as found in model.check_vs_prior.
3. model.check_vs_prior now checks the extinction law and sets the correct evaluation method
4. internally, dust_extinction works with wavenumbers in explicit units of 1/micron. We set these wavenumbers explicitly once and avoid recreating instances of the exinction model at every call, significantly accelerating runs that use Ebv.
5. We make „extinction_model“ an explicit kwarg of all LightCurveModelContainers 
6. We reduce the number of explicitly specified kwargs of LightCurveModelContainer-Children
7. SN models: rearange #425 more explicitly to tell apart salt and ccsn models, while making explicit amplitude sampling available, too.


mcoughlin:
@Hen42rik I am OK with this once the tests are green.

Hen42rik:
@mcoughlin This works out now locally, so unless I did something stupid while pushing this should be it.

mcoughlin (approved):
LGTM

PR #450, opened 2026-09-14
Delete flake/pep8 and replace with ruff, format repo using ruff

tylerbarna:
Pep8 files are outdated and can be replaced with ruff; I have also ignored a couple of nuisance rules that are present throughout the code but are arguably justified to ignore (Bare except, Ambiguous var name, Local var assigned but unused).

[318aaec](https://github.com/nuclear-multimessenger-astronomy/nmma/pull/450/commits/318aaec92083472e3c2e85fd28b818b7bb133849) expands the scope of the PR to format the repo such that all areas of the code that cause ruff to raise errors are fixed. This makes the PR seem much more significant, but the vast majority of these changes are removing trailing whitespace, fixing incorrectly formatted indents, and organizing imports, along with some multiline formatting of lists and the like. If we wish to not apply this change as part of this PR, I can revert this commit


Hen42rik (approved):
This is fully in line with the group discussion at Fahren. Thanks a lot

Hen42rik:
@mcoughlin If you second, please merge:)

mcoughlin (approved):
LGTM

PR #452, opened 2026-09-14
Run PR workflows when requesting merge into dev

tylerbarna:
Because our development workflow is such that most PRs should be merged into dev before a series of changes are collected and merged into main when releasing a new version, we should also be running checks for PRs merging into dev so merging dev into main doesn't require disentangling potentially multiple issues from several dev PRs that weren't caught beforehand.

This PR adds the dev branch to relevant workflows so the checks run when attempting to merge into dev


mcoughlin (approved):
LGTM

PR #454, opened 2026-09-14
Publish weekly unstable releases

tylerbarna:
Adds workflow to publish a weekly snapshot of the dev branch with a log of changes made; if no changes are made in a given week, a release is not generated.

This could be useful for identifying when a more complex issue might have been introduced that can't be identified based on looking at the blame of a specific file


mcoughlin (approved):
LGTM

PR #455, opened 2026-09-14
Remove outdated license

tylerbarna:
Previous PR added a reference to the GNU license in the pyproject.toml that is outdated and raises and error; this should be solved by removing the line


tylerbarna:
should be a dev pr

PR #457, opened 2026-09-14
Enforce pre-commit ruff hooks

tylerbarna:
adds pre-commit hook and docs for developers to ensure commits, by default, are required to pass ruff checks. In relevant cases, this can be bypassed with the `--no-verify` flag


mcoughlin (dismissed):
LGTM

tylerbarna:
@mcoughlin had to add `python-ligo-lw` as a dep so the unit tests stopped failing, not sure if this is a significant enough change that it should be made a separate PR (also unsure why this is an issue, as nothing I'm touching should affect the unit tests, but I digress)

mcoughlin (approved):
LGTM

PR #458, opened 2026-09-14
Split docker into build and run stages

tylerbarna:
This updates the docker file to use a [multi-stage build](https://docs.docker.com/build/building/multi-stage/), which should theoretically reduce the footprint of the container while also speeding up start time. 

I have been able to build the dockerfile on my local system


mcoughlin (approved):
LGTM

PR #459, opened 2026-09-14
Update docs to reflect new contributor guidelines

tylerbarna:
This adds some additional language to the contributing section of the docs to clarify the established development procedure. NMMA is versioned such that contributors should submit PRs to merge into the dev branch. Periodically, the dev branch will be merged into main and a new version will be released.


Hen42rik (changes requested):
This looks like a good move in the right direction, but I think it will require further modification once the new worfklows and test-pipelines are in place. I would prefer to defer merging this until the new testing is sufficiently tested.

tylerbarna:
@Hen42rik Given our discussion regarding the workflows already being in place, should we still hold off on this? I don't believe the language of the docs will necessarily change if further changes are made to the testing workflows; we will still have a policy that all checks should pass for a PR before merging it regardless of future updates to the tests

PR #460, opened 2026-09-14
Add readiness endpoint to api

tylerbarna:
adds an endpoint to the API to check if the container is ready to receive requests, distinct from the health endpoint, which determines if the api is responsive in general.


mcoughlin (approved):
LGTM

Issue #437, opened 2026-09-10
Docstrings aren't consistent/autodoc-ready; add Sphinx autodoc API reference (see jester)


ThibeauWouters:
## Summary

`doc/` currently only builds hand-written narrative guides (`doc/conf.py` enables `myst_parser`, `sphinx_copybutton`, `sphinx_github_changelog` — no `sphinx.ext.autodoc`/`sphinx.ext.napoleon`). There is no auto-generated API reference, and if we tried to add one today it wouldn't render cleanly because docstrings across the codebase aren't in consistent, autodoc-friendly shape yet.

## Current state of docstrings

- **Style is mixed.** NumPy-style (`Parameters` / `Returns` with `----------` underlines) is the dominant convention in the core physics/likelihood modules (`nmma/core/base.py`, `nmma/em/model.py`, `nmma/em/em_likelihood.py`, `nmma/eos/eos_likelihood.py`, `nmma/joint/joint_likelihood.py`, `nmma/gw/gw_likelihood.py`), but several files use Google-style `Args:`/`Returns:` blocks or reST `:param:`/`:type:` directives instead (`nmma/core/conversion.py`, `nmma/em/io.py`, `nmma/em/lightcurve_generation.py`, `nmma/em/utils.py`, `nmma/em/training.py`, `nmma/joint/generation.py`, `nmma/mlmodel/resnet.py`).
- **Coverage is partial.** Spot-checked AST-level coverage in core modules is roughly 30–50% of functions/classes (e.g. `core/conversion.py` 24/62, `em/model.py` 33/99, `eos/eos_likelihood.py` 18/47), with a lot of the rest being undocumented or one-line prose with no parameter/return documentation.
- **CLI `*_parsing.py` modules are essentially undocumented** — `em_parsing.py`, `eos_parsing.py`, `joint_parsing.py`, `gw_parsing.py` have 0–1 docstrings across dozens of argparse arguments. These modules define the user-facing CLI flags for `nmma-analysis`, `lightcurve-analysis`, etc., so this is the part that would benefit most from an API/CLI reference.
- **No docstring linting is configured** — `pyproject.toml`/`.flake8` only run plain flake8 (`C,E,F,W,B,B950`); there's no `flake8-docstrings`, `pydocstyle`, `interrogate`, or ruff `D`-rule check to keep this consistent going forward.

[...]

PR #461, opened 2026-09-15
Implement strict Sphinx check for docs build, remove stray temp files

tylerbarna:
This hardens the sphinx check on the docs to ensure any warnings on build raise a fail state while also continuing on so all warnings/errors are raised in the check. The docs files are updated to pass the sphinx check, and a handful of temp files that were erroneously version controlled have been deleted and added to .gitignore. Finally, this updates the workflow so docs-specific PRs do not run python unit-tests that are not relevant


Hen42rik:
Is this checked with the docs team, e.g. @NatalieW96 ?

tylerbarna:
@Hen42rik Checked with the docs team, this shouldn't interfere with anything they're working on, as this is a check on building the github pages docs rather than the in-code documentation.

nnarenraju:
I would still wait for Natalie to approve this. Although work was primarily done on in-code docs, there was some work put into setting up the API docs. We will check for conflicts first.

Hen42rik:
I would also appreciate input from @ThibeauWouters who had brought up some ideas for best practice in #437

Hen42rik:
I suggest you discuss this in person while everyone is still available @tylerbarna @NatalieW96 @ThibeauWouters @nnarenraju

tylerbarna:
> I would also appreciate input from @ThibeauWouters who had brought up some ideas for best practice in #437

Ah yes I agree with the points made in that issue, though I think this PR is a somewhat independent thing. This PR just enforces the existing sphinx check so the check is a proper test of the build. It also removes some random tracked temp files that shouldn't be present and skips unnecessary unit tests when a PR is only related to documentation and doesn't change actual code.

This PR is more housekeeping rather than a new feature, which should probably be contained in a separate PR(s)

ThibeauWouters:
LGTM @Hen42rik it is different

tylerbarna:
@Hen42rik Could I get a merge on this one so we don't run into some stale merge issues with subsequent workflow updates? I also noticed there's a link to docs that don't exist currently in dev; this PR also fixes that

Hen42rik (approved):

PR #463, opened 2026-09-15
Remove python-ligo-lw requirement

tylerbarna:
This was added as part of the docker overhaul, but it seems to be causing issues locally for macos users, and, upon further testing, it _seems_ the docker image can build without it. This PR removes this req; if the docker test passes, this should be fine to merge


tylerbarna:
should be into dev

PR #464, opened 2026-09-15
remove python-ligo-lw req

tylerbarna:
This was added as part of the docker overhaul, but it seems to be causing issues locally for macos users, and, upon further testing, it seems the docker image can build without it. This PR removes this req; if the unit test passes, this should be fine to merge


tylerbarna:
Yeah, for whatever reason, `python-ligo-lw` is required to pass our unit tests; this is confusing to me since presumably we've had passing unit tests for some time before I ran into the issue and added the package

tylerbarna:
Best I can figure, `gwpy` is relying on ligo-lw being imported in the scope of the script being run rather than it being imported in its own scope

tylerbarna:
@Hen42rik I think removing the specification of ligolw in the one table arg removes the error without installing python-ligo-lw from the reqs; furthermore, it seems [gwpy is no longer using python-ligo-lw anyway](https://github.com/gwpy/gwpy/blob/main/CHANGELOG.rst#id35), so I think this might be the fix? Correct me if I'm wrong, but all of these failing unit tests are caused by fiesta, right?

Update: reran the unit tests on another PR that had already passed previously and the same 7 tests failed, so yeah this unit test failing should be fiesta, at least by my eye.

Hen42rik (approved):

Hen42rik (changes requested):
Please update your base to current dev in order to make sure tests still pass, otherwise looks good

Hen42rik:
@tylerbarna Can you please update this PR to reflect the current dev branch?

Hen42rik (approved):

PR #470, opened 2026-09-18
update fiesta imports

haukekoehn:

PR #471, opened 2026-09-22
Fiesta fix

Hen42rik:
This PR makes multiple modifications
- it introduces a new CosmologyConverter-class that assures all distance/redshift conversions across nmma remain consistent and use the same cosmology.
- it adds tests for this functionality in line with the new design principles for tests that were discussed at Fahren
- it removes the black formatting from the precommit.yaml in favour of ruff's formatting
- it provides a minimal import/typo fix over #470
- it adds tests for the constants used in nmma


mcoughlin (approved):
Great!

PR #472, opened 2026-09-22
Workflow rework

thomasckng:


.github/workflows/ci.yml
```diff
@@ -1,45 +1,52 @@
-name: Unit Test nmma [ubuntu]
+name: CI
 
 on:
   push:
-    branches:
```
tylerbarna, line 5: I think this removes running workflows on dev at the moment

tylerbarna: And also removes the unstable weekly releases

